PR #2, opened 2018-08-07
GAWB-3639: Adding App Engine and Cloud Endpoints 

ansingh7115:
**Putting this up for preliminary review! Please leave feedback on larger issues. (Disregard the the minor stuff for now)**

What's I'm adding right now: 
- Making Circle CI tests pass run on PRs - this should work once I add config rendering
- Add the avram docker image to the broadinstitute repo (currently in my personal repo)
- Clean up Dockerfile and figure out what it's actually supposed to be for...
- A status endpoint which checks if the db is up
- Add a README
 

In all cases:


.circleci/config.yml
```diff
@@ -0,0 +1,43 @@
+# Scala CircleCI 2.0 configuration file
+#
+# Check https://circleci.com/docs/2.0/sample-config/ for more details
+#
+version: 2
+jobs:
+  build:
+    docker:
+      # specify the version you desire here
+      - image: ansingh7115/avram
```
ansingh7115: need to change this

deploy.sh
```diff
@@ -0,0 +1,42 @@
...
+if [ "$GIT_BRANCH" == "develop" ]; then
+	ENVIRONMENT="dev"
+elif [ "$GIT_BRANCH" == "alpha" ]; then
+        ENVIRONMENT="alpha"
+elif [ "$GIT_BRANCH" == "staging" ]; then
+	ENVIRONMENT="staging"
+elif [ "$GIT_BRANCH" == "master" ]; then
+	ENVIRONMENT="prod"
+elif [ "$GIT_BRANCH" == "gawb3639" ]; then
+	ENVIRONMENT="dev"
```
ansingh7115: deploying my branch to dev for now to make sure it works - will remove this before merging

deploy.sh
```diff
@@ -0,0 +1,52 @@
...
+fi
+
+GOOGLE_PROJECT=broad-avram-$ENVIRONMENT
+SERVICE_VERSION=version1
+
+# pull the credentials for the service account
+docker run --rm -e VAULT_TOKEN=$VAULT_TOKEN broadinstitute/dsde-toolbox vault read --format=json "secret/dsde/avram/$ENVIRONMENT/deploy-account.json" | jq .data > deploy_account.json
+
+# Auth with deploy service account
+gcloud auth activate-service-account --key-file=deploy_account.json
```
breilly2: Is it okay for us to be running `gcloud` commands like this or should we be doing it inside a docker container like bond does? https://github.com/DataBiosphere/bond/blob/develop/deploy.sh#L33
Attn: @jacmrob

deploy.sh
```diff
@@ -0,0 +1,52 @@
...
+  -e GOOGLE_PROJ=$GOOGLE_PROJECT \
+  -e SERVICE_VERSION=$SERVICE_VERSION \
+  -e INPUT_PATH=/app \
+  -e OUT_PATH=/app \
+  -e VAULT_TOKEN=$VAULT_TOKEN \
+  -e ENVIRONMENT=$ENVIRONMENT \
+  broadinstitute/dsde-toolbox render-templates.sh
+
+# get appenginge sdk
+curl 'https://storage.googleapis.com/appengine-sdks/featured/appengine-java-sdk-1.9.64.zip' > /tmp/appengine.zip
```
breilly2: We can probably use the avram docker image which already has the App Engine SDK.

Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM circleci/openjdk:8-jdk
+
+
+USER root
+
+RUN wget https://dl.google.com/dl/cloudsdk/channels/rapid/google-cloud-sdk.zip && unzip google-cloud-sdk.zip && rm google-cloud-sdk.zip
```
breilly2, line 6: The `sbt-appengine` plugin only needs the App Engine SDK, not the Google Cloud SDK, for building. However, we may want to keep the Google Cloud SDK in this image if we start using it from `deploy.sh`.

Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+RUN google-cloud-sdk/install.sh --usage-reporting=true --path-update=true --bash-completion=true --rc-path=/.bashrc --additional-components app cloud-datastore-emulator
+ENV PATH /google-cloud-sdk/bin:$PATH
+
+RUN gcloud components update --quiet && gcloud components install beta --quiet
+
+RUN mkdir -p /home/gcloud \
+&& curl 'https://storage.googleapis.com/appengine-sdks/featured/appengine-java-sdk-1.9.63.zip' > /tmp/appengine.zip \
```
breilly2: Should be using `1.9.64`. I wonder if we can externalize the version from `Dependencies.scala` so we can configure both that and this from one place.

project/Dependencies.scala
```diff
@@ -26,24 +24,27 @@ object Dependencies {
   val jacksonAnnotations: ModuleID = "com.fasterxml.jackson.core" % "jackson-annotations" % jacksonV
   val jacksonDatabind: ModuleID =    "com.fasterxml.jackson.core" % "jackson-databind"    % jacksonV
   val jacksonCore: ModuleID =        "com.fasterxml.jackson.core" % "jackson-core"        % jacksonV
+  val jacksonScalaModule: ModuleID = "com.fasterxml.jackson.module" %% "jackson-module-scala" % "2.9.6"
+  val lombok: ModuleID = "org.projectlombok" % "lombok" % "1.18.2" % "provided"
```
breilly2: I think we don't need lombok because of conveniences that scala gives us.

openapi.yaml
```diff
@@ -0,0 +1,53 @@
+# Copyright 2017 Google Inc. All Rights Reserved.
```
breilly2: Is this a generated file or are we making changes manually? This looks like the same information as in the `@Api` annotations so it would be preferable to maintain it in only one place.

ansingh7115: It would definitely be awesome if we could autogenerate this! I will probably leave that for future work though

deploy.sh
dmohs: Sorry to butt-in on a random PR, but you might think about whether your needs here are unique. If not, this might be a nice feature to build-in for the team. For example, you optionally read a `.deploy` file that contains this option and instruct devs to create one (which is in the `.gitignore`) to get this behavior. Just a thought—take with a grain of salt.

src/main/scala/org/broadinstitute/dsde/workbench/avram/util/Logger.scala
```diff
@@ -0,0 +1,34 @@
...
+  private def entry(text: String): LogEntry = LogEntry
+    .newBuilder(StringPayload.of(text))
+    .setSeverity(Severity.INFO)
+    .setLogName(logName)
+    .setResource(MonitoredResource.newBuilder(resourceType).setLabels(Map("project_id" -> projectId, "app" -> "avram").asJava).build)
+    .build
+
+  def log(message: String) = {
+    logging.write(Collections.singleton(entry(message)))
+  }
```
ahaessly: we might want the log method to take the severity also. we would also need to add that param to entry

src/main/webapp/WEB-INF/logging.properties
```diff
@@ -0,0 +1,13 @@
...
+# To use this configuration, copy it into your application's WEB-INF
+# folder and add the following to your appengine-web.xml:
+# 
+# <system-properties>
+#   <property name="java.util.logging.config.file" value="WEB-INF/logging.properties"/>
+# </system-properties>
+#
+
+# Set the default logging level for all loggers to WARNING
+.level = INFO
```
ahaessly: comment doesn't match setting.

src/main/webapp/WEB-INF/web.xml
```diff
@@ -0,0 +1,22 @@
...
+            <param-name>restricted</param-name>
+            <param-value>false</param-value>
+        </init-param>
+    </servlet>
+    <servlet-mapping>
+        <servlet-name>EndpointsServlet</servlet-name>
+        <url-pattern>/*</url-pattern>
+    </servlet-mapping>
+    <welcome-file-list>
+        <welcome-file>index.html</welcome-file>
```
ahaessly: do we have to have this?

ahaessly (approved):
just a few comments - otherwise 👍

src/main/webapp/WEB-INF/logging.properties
dmohs: Nice. Suggested comment: "Set the default logging level for all loggers." to avoid the mismatch altogether.

Dockerfile
breilly2: Greg Polumbo and I chatted about docker images for independent services. It may be that we want to have 2: one for building (used by CircleCI) and one for running (used by FiaB). We can probably prune this image a little but I say we wait until we make the image for FiaB which may want some of the pieces in here.

Also, https://cloud.google.com/sdk/docs/quickstart-debian-ubuntu has some suggestions for installing the SDK into a docker image.

Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM circleci/openjdk:8-jdk
+
+
+USER root
+
+RUN wget https://dl.google.com/dl/cloudsdk/channels/rapid/google-cloud-sdk.zip && unzip google-cloud-sdk.zip && rm google-cloud-sdk.zip
+RUN google-cloud-sdk/install.sh --usage-reporting=true --path-update=true --bash-completion=true --rc-path=/.bashrc --additional-components app cloud-datastore-emulator
+ENV PATH /google-cloud-sdk/bin:$PATH
+
+RUN gcloud components update --quiet && gcloud components install beta --quiet
```
breilly2: Do we need `beta` for anything? If not, we should remove that to avoid accidentally relying on beta software. If so, we should add a comment about what we need so we can keep track of those dependencies and eventually remove `beta`.

ansingh7115: I think that was me copy pasting stuff :P

deploy.sh
breilly2: It's a nice idea, but it's not isolated to here. If running this script from a local workstation, one can just pass in `develop` or whatever is appropriate. The need for recognizing other branches is for triggering deploys from Jenkins which also requires the branch to be listed here: https://github.com/broadinstitute/dsp-jenkins/blob/master/src/main/resources/IndieDeployConfig.conf.

deploy.sh
breilly2: I checked with Cloud Accounts and Matt Bemis said that using docker is preferable to avoid interference between jobs. I am working on the conversion.

deploy.sh
```diff
@@ -0,0 +1,43 @@
...
+# Auth with deploy service account
+gcloud auth activate-service-account --key-file=deploy_account.json
+
+# get appenginge sdk
+curl 'https://storage.googleapis.com/appengine-sdks/featured/appengine-java-sdk-1.9.64.zip' > /tmp/appengine.zip
+unzip /tmp/appengine.zip
+export APPENGINE_SDK_HOME=$PWD/appengine-java-sdk-1.9.64
+
+# deploy endpoints
+gcloud endpoints services deploy $PWD/openapi.yaml --project=$GOOGLE_PROJECT
```
breilly2: Is there a reason for deploying the endpoints first? In my mind, the endpoints are built on top of the App Engine app so it seems more logical to deploy the app first and then the endpoints.

openapi.yaml
breilly2: Let's at least get rid of the Google copyright and license. We can replace with our own if we want.

project/Dependencies.scala
```diff
@@ -1,8 +1,6 @@
 import sbt._
 
 object Dependencies {
```
breilly2, line 3: I'd like to try to trim down the dependencies as much as possible. For example, I question whether we need `logbackClassic`, `swaggerUi`, `httpClient`, `unfilteredFilter`, and `googleRpc`, and we should make sure we're using `ficus` and `enumeration` if we're going to keep those.

src/main/resources/META-INF/jdoconfig.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
```
breilly2: We don't need this because we're not using JDO.

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
```diff
@@ -1,82 +1,20 @@
...
-        case other =>
-          Future.successful(LogEntry(s"$other", Logging.DebugLevel)) // I don't really know when this case happens
-      }
-      entry.map(_.logTo(logger))
-    }
-
-    DebuggingDirectives.logRequestResult(LoggingMagnet(log => myLoggingFunction(log)))
+  @ApiMethod(name = "ping", httpMethod = "get", path = "ping")
+  def ping: Pong = {
+    log.log("HIII")
```
breilly2: HI!!1

src/main/scala/org/broadinstitute/dsde/workbench/avram/util/Logger.scala
breilly2: Google recommends using `java.util.logging.Logger` (https://cloud.google.com/appengine/docs/standard/java/logs/) which I think we should do. Adopting that interface gives us `logger.info()`, `logger.warn()`, `logger.info()`, etc.

src/main/webapp/WEB-INF/appengine-web.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<appengine-web-app xmlns="http://appengine.google.com/ns/1.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://kenai.com/projects/nbappengine/downloads/download/schema/appengine-web.xsd appengine-web.xsd">
+    <application>broad-avram-dev</application>
```
breilly2, line 3: We'll need to make this dynamic at some point. Probably when we tackle config rendering.

src/main/webapp/WEB-INF/appengine-web.xml
```diff
@@ -0,0 +1,10 @@
...
+    <version>1</version>
```
breilly2, line 4: Not sure how to handle this. Jenkins deploys should probably deploy to a real version. It would be nice if a local checkout did not have this version configured so we don't accidentally deploy WIP to that real version. I've been editing this locally to something like `gawb-3691` and so far have not forgotten to do that (I think) but it makes me nervous.

deploy.sh
ansingh7115: I did that way partly because Bond did it that way and so I kept the ordering - but you're right.

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
ansingh7115: Haha, testing the logging - I'll remove it

src/main/scala/org/broadinstitute/dsde/workbench/avram/util/Logger.scala
ansingh7115: I attempted to use `java.util.logging.Logger` but for some reason was not able to see the log messages in Stackdriver. I think I might be doing something wrong.

deploy.sh
ansingh7115: That's a good point - going forward, we would definitely want to change and test the deploy script and do that in an easy, reasonable way. I'll create a ticket and prioritize it to make sure we consider that!

src/main/webapp/WEB-INF/web.xml
ansingh7115: no - left over from the prototype version which had a ui - I'll remove it

.circleci/config.yml
```diff
@@ -0,0 +1,43 @@
...
+          - v1-dependencies-{{ checksum "build.sbt" }}
+          # fallback to using the latest cache if no exact match is found
+          - v1-dependencies-
+
+      - run: cat /dev/null | sbt test:compile
+
+      - save_cache:
+          paths:
+            - ~/.m2
+          key: v1-dependencies--{{ checksum "build.sbt" }}
```
breilly2: Just reading up on save_cache (https://circleci.com/docs/2.0/configuration-reference/#save_cache)... we should use the checksum of `project/Dependencies.scala`.

deploy.sh
```diff
@@ -0,0 +1,44 @@
...
+	ENVIRONMENT="staging"
+elif [ "$GIT_BRANCH" == "master" ]; then
+	ENVIRONMENT="prod"
+else
+	echo "Unknown Git branch $GIT_BRANCH"
+	#exit 1
+fi
+
+GOOGLE_PROJECT=broad-avram-$ENVIRONMENT
+SERVICE_VERSION=version1
```
breilly2: What is this used for? Does this need to be something other than a fixed value?

breilly2 (approved):
👍 I added a couple more comments just now, but I think there's more value getting this merged than spending more time tweaking minor things.

deploy.sh
ansingh7115: I'm removing the version name for now (I just adopted it from Bond) but we'll probably want versioned APIs at some point in the future

ansingh7115:
When I ran the deploy again after pushing the image manually (without the line in the deploy script that builds the image), the deploy used an earlier image and failed. It didn't pull the latest image. I'm wondering if we should have an explicit `docker pull` in the deploy script.

Also, I think most other services have building and pushing their docker images to dockerhub as part of their automated build/deploy process

PR #1, opened 2018-08-06
Initial database setup (proof-of-concept)

breilly2:
This will get us started. Still need to get a connection pool set up. Also need to make the configs dynamic but I figure that will come with setting up build automation and more deployment environments. Cloud SQL instance, database, and user (all named "avram") have been created in the broad-avram-dev project. Passwords are in vault.

A working build is deployed at https://gawb3691-dot-broad-avram-dev.appspot.com/

Have you read [CONTRIBUTING.md](../CONTRIBUTING.md) lately? If not, do that first.

I, the developer opening this PR, do solemnly pinky swear that:

- [ ] I've documented my API changes in Swagger

In all cases:

- [ ] Get two thumbsworth of review and PO signoff if necessary
- [ ] Verify all tests go green
- [ ] Squash and merge; you can delete your branch after this
- [ ] Test this change deployed correctly and works on dev environment after deployment

PR #4, opened 2018-08-13
GAWB-3691 Initial database setup

breilly2:
(Replaces #1)

This will get us started. Still need to get a connection pool set up. Also need to make the configs dynamic but I figure that will come with setting up build automation and more deployment environments. Cloud SQL instance, database, and user (all named "avram") have been created in the broad-avram-dev project. Passwords are in vault.

A working build is deployed at https://gawb3691-dot-broad-avram-dev.appspot.com/. Try https://gawb3691-dot-broad-avram-dev.appspot.com/avram/v1/now

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
@@ -4,7 +4,7 @@ import java.sql.{Connection, DriverManager, PreparedStatement, ResultSet}
 
 import com.google.api.server.spi.config.{Api, ApiMethod}
 import com.typesafe.config.ConfigFactory
-import org.broadinstitute.dsde.workbench.avram.util.Logger
+import org.broadinstitute.dsde.workbench.avram.util.{DataSource, Logger}
 
 import scala.concurrent.duration.Duration
 import scala.concurrent.{Await, Future}
@@ -15,6 +15,7 @@ import scala.beans.BeanProperty
 
 case class Pong()
 case class Now(@BeanProperty message: String)
+case class DbPoolStats(@BeanProperty numActive: Int, @BeanProperty numIdle: Int, @BeanProperty totalConnections: Int)
 
 @Api(name = "avram", version = "v1", scopes = Array("https://www.googleapis.com/auth/userinfo.email"), clientIds = Array(""), audiences = Array(""))
 class AvramRoutes {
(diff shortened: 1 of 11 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
```diff
@@ -13,4 +25,35 @@ class AvramRoutes {
...
+  }
+
+  private def fetchTimestampFromDBWithSlick(): String = {
+    val now = Await.result(DataSource.database.run(sql"select now()".as[String]), Duration.Inf)
+    now.head
+  }
+
+  private def fetchTimestampFromDBWithJDBC(): String = {
+    val config = ConfigFactory.load().getConfig("postgres")
+    val conn: Connection = DriverManager.getConnection(config.getString("url"), config.getString("user"), config.getString("password"))
```
ahaessly: why doesn't this use the datasource?

src/test/resources/application.conf
```diff
@@ -0,0 +1,8 @@
+postgres = {
```
ahaessly: will this and the other application.conf eventually be removed from this repo and generated from firecloud-develop?

ansingh7115:
Circle CI tests are failing with the following message:

`Cause: java.io.IOException: The Application Default Credentials are not available. They are available if running on Google App Engine, Google Compute Engine, or Google Cloud Shell. Otherwise, the environment variable GOOGLE_APPLICATION_CREDENTIALS must be defined pointing to a file defining the credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.`

The page in the error message: https://cloud.google.com/docs/authentication/production#providing_credentials_to_your_application
is actually pretty useful. Setting the  GOOGLE_APPLICATION_CREDENTIALS  in the tests would get rid of this issue

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
```diff
@@ -1,9 +1,21 @@
...
+
+import scala.concurrent.duration.Duration
+import scala.concurrent.{Await, Future}
+import scala.concurrent.ExecutionContext.Implicits.global
+import slick.jdbc.PostgresProfile.api._
+
+import scala.beans.BeanProperty
 
 case class Pong()
+case class Now(@BeanProperty message: String)
```
breilly2, line 14: I agree with what Anu said in person that it's annoying to have to annotate all properties with `@BeanProperty`. It's unclear to me what exactly is insisting that these objects have getters/setters.

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
breilly2: I should delete this. It's left over from when I thought I was seeing some slick-specific problems.

src/test/resources/application.conf
breilly2: No. As an independent service its config belongs here, not in `firecloud-develop`.

There's still an open question about how to configure independent services for deployment in FiaB. Options are maintaining copies in `firecloud-develop` or having FiaB config scripts pull config templates from independent service repos. Greg Polumbo has been working on getting Martha and Bond into FiaB and is, for now, maintaining separate copies in `firecloud-develop` just because it was the easiest to implement.

breilly2:
Right. The JDBC URL in the config is for the Avram dev Cloud SQL database which requires credentials to access from outside of the Google Cloud ecosystem. We could try to pull in these credentials but builds would be sharing the Cloud SQL instance which is probably not best. Alternately, we could ask Circle CI for a PostgreSQL DB and use a different JDBC URL for the tests. I think that's the better option so I'll work on setting that up.

ansingh7115 (approved):
Aside from the Circle CI tests failing, looks awesome! We should start doing things in Transactions (I think you said this face to face), but we can do that next iteration. Really glad we can use Slick and that connection pooling is a go. :)

breilly2:
Superseded by #5

PR #5, opened 2018-08-28
Database setup

breilly2:
#4 rebased onto develop

In all cases:

### DIFF
--- a/docker/run-postgres.sh
+++ b/docker/run-postgres.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# The CloudSQL console simply states "PostgreSQL 5.6" so we may not match the minor version number
+# The CloudSQL console simply states "PostgreSQL 9.6" so we may not match the minor version number
 POSTGRES_VERSION=9.6
 start() {
     echo "attempting to remove old $CONTAINER container..."
--- a/docker/run-postgres.sh
+++ b/docker/run-postgres.sh
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+
+# The CloudSQL console simply states "PostgreSQL 5.6" so we may not match the minor version number
+POSTGRES_VERSION=9.6
+start() {
+    echo "attempting to remove old $CONTAINER container..."
+    docker rm -f $CONTAINER || echo "docker rm failed. nothing to rm."
+
+    # start up PostgreSQL
+    echo "starting up postgres container..."
+    docker run --name $CONTAINER -e POSTGRES_USER=avram -e POSTGRES_PASSWORD=test -e POSTGRES_DB=testdb -d -p 5432:5432 circleci/postgres:$POSTGRES_VERSION-alpine-ram
+}
+
+stop() {
+    echo "Stopping docker $CONTAINER container..."
+    docker stop $CONTAINER || echo "$CONTAINER stop failed. container already stopped."
+    docker rm -v $CONTAINER || echo "$CONTAINER rm -v failed.  container already destroyed."
+}
+
+CONTAINER=postgres-avram
+COMMAND=$1
+
+if [ ${#@} == 0 ]; then
+    echo "Usage: $0 stop|start"
+    exit 1
+fi
+
+if [ $COMMAND = "start" ]; then
+    start
+elif [ $COMMAND = "stop" ]; then
+    stop
+else
+    exit 1
+fi
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -1,11 +1,14 @@
 import sbt._
 
 object Dependencies {
-  val jacksonV      = "2.9.5"
-  val googleV       = "1.23.0"
-  val scalaLoggingV = "3.9.0"
-  val scalaTestV    = "3.0.5"
-  val slickV        = "3.2.3"
+  val jacksonV        = "2.9.5"
+  val googleV         = "1.23.0"
+  val scalaLoggingV   = "3.9.0"
+  val scalaTestV      = "3.0.5"
+  val slickV          = "3.2.3"
+  val postgresDriverV = "42.2.4"
+  val socketFactoryV  = "1.0.10"
+  val dbcpV           = "2.5.0"
 
   val workbenchUtilV    = "0.3-0e9d080"
   val workbenchModelV   = "0.11-2ce3359"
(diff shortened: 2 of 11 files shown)


ahaessly (approved):
👍

docker/run-postgres.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+
+# The CloudSQL console simply states "PostgreSQL 5.6" so we may not match the minor version number
```
rtitle: you mean 9.6? :)

project/Dependencies.scala
```diff
@@ -59,10 +62,12 @@ object Dependencies {
   val sam: ModuleID = "org.broadinstitute.dsde.sam-client" %% "sam" % samV
 
   val slick: ModuleID =     "com.typesafe.slick" %% "slick"                 % slickV
-  val hikariCP: ModuleID =  "com.typesafe.slick" %% "slick-hikaricp"        % slickV
-  val mysql: ModuleID =     "mysql"               % "mysql-connector-java"  % "8.0.11"
+  val dbcp2: ModuleID = "org.apache.commons" % "commons-dbcp2" % dbcpV
```
rtitle, line 65: Is dbcp a connection pooling library? Did we confirm that this is OK in Google App Engine where threading is more controlled?

src/main/resources/application.conf
```diff
@@ -0,0 +1,14 @@
...
+// along with properties for use with a slick Database
+// Note: this is _not_ appropriate for passing to a slick Database factory method
+dbcpDataSource = {
+  driverClassName = org.postgresql.Driver
+  url = "jdbc:postgresql://google/avram?ssl=false&socketFactoryArg=broad-avram-dev:us-central1:avram&socketFactory=com.google.cloud.sql.postgres.SocketFactory"
+  username = "avram"
+  password = ""
+  maxTotal = 20 // http://slick.lightbend.com/doc/3.2.3/database.html
+  slick = {
+    numThreads = 10 // https://github.com/brettwooldridge/HikariCP/wiki/About-Pool-Sizing
```
rtitle, line 11: Might want to adjust these comments since we're not using HikariCP. Are `numThreads`, `queueSize` still relevant?

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
```diff
@@ -16,4 +25,26 @@ class AvramRoutes {
   def ping: Pong = {
     Pong()
   }
+
+  // TODO: remove this endpoint when we have more meaningful ways to test database queries
+  @ApiMethod(name = "now", httpMethod = "get", path = "now")
+  def now: Now = {
+    // Explicitly use a Future to make sure the implicit ExecutionContext is being used
```
rtitle, line 32: So `scala.concurrent.ExecutionContext.Implicits.global` works in App Engine? I was wondering if we needed to define our own single-threaded ExecutionContext.

rtitle: Also I know this is a toy API but isn't this essentially doing:
```
Await.result(Future(Await.result(database.run(...))))
```
? Not sure you need the outer `Await`/`Future`.

docker/run-postgres.sh
breilly2: Oops! I thought I had caught everything related to MySQL. I guess I missed one.

src/test/scala/org/broadinstitute/dsde/workbench/avram/integration/PostgreSQLSpec.scala
```diff
@@ -0,0 +1,22 @@
...
+
+import scala.concurrent.Await
+import scala.concurrent.duration.Duration
+
+class PostgreSQLSpec extends FreeSpec {
+
+  "should fetch" - {
+    "using slick" in {
+      val db = DataSource.database
+      try {
```
rtitle, line 15: In the future it might be good to add some unit test utilities for database stuff. For example Leo has this trait:

https://github.com/DataBiosphere/leonardo/blob/develop/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/db/TestComponent.scala

which lets you write tests like 
```
it should "foo" in isolatedDbTest {
   val result = dbFutureValue { someQuery }
   result shouldBe 42
}
```

rtitle:
Looks good to me! Just some questions above. It'll be cool to see some postgres-json stuff in the future.

project/Dependencies.scala
breilly2: Yes, DBCP is a connection pooling library. I had better luck with it in my simple experiments than with HikariCP. We should do some more stress testing to be sure, but there are reports of others successfully using DBCP on app engine standard.

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
breilly2: My understanding of threads in app engine is that _background_ threads aren't allowed in app engine standard with automatic scaling (though _are_ allowed with basic or manual scaling, both of which do not scale to zero). Threads during requests are supported. You can use either the GAE thread API or the standard Java management API, though you can't call other GCP API calls from Java-managed threads. If we find we need to do that, that's when we'll need a custom execution context that uses the GAE thread API.

breilly2: Ah, you're right. Looks like I double wrapped it.

src/main/resources/application.conf
breilly2: These are being used to configure DBCP. See DataSource.scala. The documentation link talks about choosing values and isn't really specific to HikariCP.

src/test/scala/org/broadinstitute/dsde/workbench/avram/integration/PostgreSQLSpec.scala
breilly2: Absolutely. I can imagine having tests that talk to postgres and tests that talk to an in-memory database for isolation depending on what exactly we're testing. There's a lot of room to explore the right mix of test environments.

project/Dependencies.scala
breilly2: I also need to check in with our GCP support team. They were supposed to be getting back to me with more specific advice.

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
breilly2: Actually, explicitly referencing a `Future` here necessitates an `ExecutionContext` in a way that `Await.result()` does not. I need to understand this better to simplify it so for now I'm leaving it as-is.

PR #6, opened 2018-08-30
GAWB-3766: Add config rendering

ansingh7115:

In all cases:


ansingh7115:
@breilly2 What db related configs do we need to add?

app.conf.ctmpl
```diff
@@ -0,0 +1,12 @@
+{{with $environment := env "ENVIRONMENT"}}
+{{with $client := vault (printf "secret/dsde/avram/%s/oauth_client" $environment)}}
+{{with $googleProject := env "GOOGLE_PROJ"}}
+{{with $serviceVersion := env "SERVICE_VERSION"}}
+
+avram {
+    serviceVersion = "{{$serviceVersion}}"
```
ansingh7115: Just stuck configy-type stuff in here to make sure pulling from vault and reading environment variables works - these aren't the final things I want to put in here. I'm wondering what else we need in config right now? I assume a db config?

rtitle: Yeah probably need a slick/postgres config like: http://slick.lightbend.com/doc/3.2.0/database.html

src/main/scala/org/broadinstitute/dsde/workbench/avram/config/AvramConfig.scala
```diff
@@ -0,0 +1,5 @@
+package org.broadinstitute.dsde.workbench.avram.config
+
+case class AvramConfig(serviceVersion: String,
```
rtitle: It'd be cool to expose the service version in some /version API endpoint. However I wonder if we could set the version in the SBT build file instead of `avram.conf`.

src/main/scala/org/broadinstitute/dsde/workbench/avram/config/AvramConfig.scala
```diff
@@ -0,0 +1,5 @@
...
+                        clientId: String,
+                        googleProject: String)
```
rtitle: Are these just examples or do you envision using `clientId` and `googleProject` in the future? (and if so, how?)

rtitle (approved):
The config infrastructure looks good. Do we have a Cloud SQL DB set up yet? If so it'd be cool to plug in the DB configs to `avram.conf` and verify the connection to Cloud SQL works.

src/main/scala/org/broadinstitute/dsde/workbench/avram/config/AvramConfig.scala
ahaessly: 👍

ahaessly (approved):

src/main/scala/org/broadinstitute/dsde/workbench/avram/config/AvramConfig.scala
ansingh7115: they're examples for now - making sure we can pull secrets and env. variables!

breilly2:
What command should I run to render config for local development?

app.conf.ctmpl
```diff
@@ -0,0 +1,21 @@
+{{with $environment := env "ENVIRONMENT"}}
+{{with $dataSource := vault (printf "secret/dsde/avram/%s/dbcDataSource" $environment)}}
```
breilly2: You missed the "p" if you were going for "dbcpDataSource" in the vault key. However, there isn't anything DBCP-specific that we're putting in vault so I think this should just be "dataSource" or "db" or something like that.

app.conf.ctmpl
```diff
@@ -0,0 +1,21 @@
...
+{{with $googleProject := env "GOOGLE_PROJ"}}
+{{with $serviceVersion := env "SERVICE_VERSION"}}
+
+// Properties for configuring an Apache DBCP BasicDataSource
+// along with properties for use with a slick Database
+// Note: this is _not_ appropriate for passing to a slick Database factory method
+dbcpDataSource = {
+  driverClassName = org.postgresql.Driver
+  url = "{{$dataSource.Data.url}}"
```
breilly2: This isn't secret.

app.conf.ctmpl
```diff
@@ -0,0 +1,21 @@
+{{with $environment := env "ENVIRONMENT"}}
+{{with $dataSource := vault (printf "secret/dsde/avram/%s/dbcDataSource" $environment)}}
+{{with $googleProject := env "GOOGLE_PROJ"}}
```
breilly2: `googleProject` and `serviceVersion` are unused here.

src/main/resources/application.conf
```diff
@@ -4,8 +4,6 @@
 dbcpDataSource = {
   driverClassName = org.postgresql.Driver
   url = "jdbc:postgresql://google/avram?ssl=false&socketFactoryArg=broad-avram-dev:us-central1:avram&socketFactory=com.google.cloud.sql.postgres.SocketFactory"
-  username = "avram"
```
breilly2: Can this file go away completely if everything is in `app.conf` now?

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
```diff
@@ -20,6 +22,10 @@ case class DbPoolStats(@BeanProperty numActive: Int, @BeanProperty numIdle: Int,
 class AvramRoutes {
 
   private val log = Logger.getLogger(getClass.getName)
+  val configFactory = ConfigFactory.parseResources("app.conf") //s.withFallback(ConfigFactory.load())
```
breilly2: We should extract this into a `Config` object so that individual components don't have to repeat the knowledge of where to read configs from.

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
```diff
@@ -20,6 +22,10 @@ case class DbPoolStats(@BeanProperty numActive: Int, @BeanProperty numIdle: Int,
...
+  private val dbcpDataSourceConfig = configFactory.as[DbcpDataSourceConfig]("dbcpDataSource")
```
breilly2: This should move into `DataSource`.

src/main/scala/org/broadinstitute/dsde/workbench/avram/config/DbcpDataSourceConfig.scala
```diff
@@ -0,0 +1,9 @@
+package org.broadinstitute.dsde.workbench.avram.config
+
+case class DbcpDataSourceConfig(driverClassName: String,
```
breilly2: Nice. I like making this more explicit. Can we have this case class and its `ValueReader` in the same file?

src/test/scala/org/broadinstitute/dsde/workbench/avram/CommonTestData.scala
```diff
@@ -14,6 +17,11 @@ import scala.concurrent.ExecutionContext
 
 trait CommonTestData { this: ScalaFutures =>
 
+  val configFactory = ConfigFactory.parseResources("app.conf").withFallback(ConfigFactory.load())
```
breilly2: I think all of this will go away if we move things around a little as I suggested above. That way there isn't any chance that config loading will diverge between main and test.

src/test/scala/org/broadinstitute/dsde/workbench/avram/integration/PostgreSQLSpec.scala
```diff
@@ -1,17 +1,19 @@
...
 import org.broadinstitute.dsde.workbench.avram.util.DataSource
 import org.scalatest.FreeSpec
+import org.scalatest.concurrent.ScalaFutures
 import slick.jdbc.PostgresProfile.api._
 
 import scala.concurrent.Await
 import scala.concurrent.duration.Duration
 
-class PostgreSQLSpec extends FreeSpec {
+class PostgreSQLSpec extends FreeSpec with CommonTestData with ScalaFutures {
```
breilly2: Why is `ScalaFutures` needed?

app.conf.ctmpl
ansingh7115: Yep - I'm removing those - I was using those to test config rendering before I rebased on top of your stuff

src/main/resources/application.conf
ansingh7115: Probs

src/main/scala/org/broadinstitute/dsde/workbench/avram/config/DbcpDataSourceConfig.scala
ansingh7115: We could - I was following the Leo and Sam format

breilly2: They're already tightly coupled because they enumerate the same fields so I think they belong as close together as possible.

PR #7, opened 2018-09-07
GAWB-3765: Initial stuff to talk to sam for authn.

breilly2:
PR for early feedback. Not done yet but please review. Thanks!

In all cases:


project/Settings.scala
```diff
@@ -42,7 +42,8 @@ object Settings {
       organization  := "org.broadinstitute.dsde.workbench",
       scalaVersion  := "2.12.6",
       resolvers ++= commonResolvers,
-      scalacOptions ++= commonCompilerSettings
+      scalacOptions ++= commonCompilerSettings,
+      addCompilerPlugin("org.scalamacros" % "paradise" % "2.1.1" cross CrossVersion.full)
```
breilly2: I don't know if this is the right place to put this. Any sbt experts out there?

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
```diff
@@ -26,6 +28,22 @@ class AvramRoutes {
     Pong()
   }
 
+  @ApiMethod(name = "authPing", httpMethod = "get", path = "authPing")
+  def authPing(r: HttpServletRequest): Pong = {
+    getToken(r) match {
```
breilly2, line 37: Ultimately, I want to make the "routes" layer extremely thin and handle token checks pretty much automatically. I learned some tricks in Scala with Cats that we might be able to use to make the code not even compile if we forget to handle auth (which will hopefully make Bernick happy).

src/main/scala/org/broadinstitute/dsde/workbench/avram/dataaccess/HttpSamDao.scala
```diff
@@ -0,0 +1,34 @@
...
+import io.circe.generic.auto._
+import io.circe.parser._
+import org.broadinstitute.dsde.workbench.avram.util.{ErrorResponse, RestClient}
+
+class HttpSamDao extends SamDao with RestClient {
+  private val log: Logger = Logger.getLogger(getClass.getName)
+  // TODO: read from config
+  val samUrl = "https://sam.dsde-dev.broadinstitute.org"
+
+  override def getUserStatus(token: String): Either[ErrorResponse, UserInfo] = {
```
breilly2: Attempting to delay throwing exceptions until absolutely necessary.

src/main/scala/org/broadinstitute/dsde/workbench/avram/util/ErrorResponse.scala
```diff
@@ -0,0 +1,36 @@
...
+case class ErrorResponse(statusCode: Int, message: String) {
+  private val log = Logger.getLogger(getClass.getName)
+
+  /**
+    * Create a Google Cloud Endpoints Framework exception matching the status code.
+    */
+  def exception: Exception = {
+    log.info(message)
+    statusCode match {
+      case 400 => new BadRequestException(message)
```
breilly2, line 24: Might make (or find) a `StatusCode` type to avoid having to remember the numbers, and maybe reduce the code here.

src/main/scala/org/broadinstitute/dsde/workbench/avram/util/RestClient.scala
```diff
@@ -0,0 +1,27 @@
...
+
+import com.softwaremill.sttp.{HttpURLConnectionBackend, Request, Uri, sttp}
+
+/**
+  * Helpful stuff for making rest calls using sttp (https://github.com/softwaremill/sttp).
+  */
+trait RestClient {
+  private val log: Logger = Logger.getLogger(getClass.getName)
+
+  implicit val sttpBackend = HttpURLConnectionBackend()
```
breilly2: This is a synchronous backend. There are many others available: https://sttp.readthedocs.io/en/latest/backends/summary.html. Suggestions welcome.

src/main/scala/org/broadinstitute/dsde/workbench/avram/dataaccess/HttpSamDao.scala
```diff
@@ -0,0 +1,34 @@
...
+    val request = buildAuthenticatedGetRequest(samUrl, "/register/user/v1?userDetailsOnly=true", token)
+    val response: Id[Response[String]] = request.send()
```
breilly2: I was super verbose here with types and intermediate vars just to keep everything straight. I definitely want to reduce the verbosity here. I think sttp has a way of plugging in serialization/deserialization so we don't have to do that explicitly. There may also be some way to consolidate the range of types used in the `Left`s so we can simply `map` instead of `match`.

src/main/scala/org/broadinstitute/dsde/workbench/avram/dataaccess/HttpSamDao.scala
```diff
@@ -0,0 +1,34 @@
+package org.broadinstitute.dsde.workbench.avram.dataaccess
+
+import java.util.logging.Logger
+
+import com.softwaremill.sttp._
+import io.circe.Decoder.Result
+import io.circe.{DecodingFailure, Json, ParsingFailure}
+import io.circe.generic.auto._
```
breilly2, line 8: Oh, BTW, this is really cool! Note the lack of encode/decode for , or annotations on, `UserInfo`.

project/Settings.scala
rtitle: I think this is the correct place for compiler plugins, in project settings. Dare I ask why you need macro paradise? :) Personally I try and steer clear of using Scala macros, but I'm not a macro expert.

breilly2: I'm glad you asked. It was for the semi-auto encoder/decoder generation in circe. However, unless it becomes a problem, I'd like to use full auto which might not require macros. I'll see if I can do away with this.

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
rtitle: Cool -- what kind of tricks are you thinking? Since auth will be common to every API it'd be nice to encapsulate it into some kind of helper function.

I'm just spitballing here:
```
def extractUserInfo(r: HttpServletRequest): Either[ErrorResponse, UserInfo] = ???

@ApiMethod(...)
def authPing(r: HttpServletRequest): Pong = {
  val responseOrError = extractUserInfo(r).map { userInfo =>
    log.info(userInfo)
    Pong()
  }
  // maybe this "get or throw" fold is common enough to make a helper function too
  responseOrError.fold {
    error => throw error.exception,
    result => result
  }
}
```

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
```diff
@@ -43,6 +61,12 @@ class AvramRoutes {
     Await.result(result, Duration.Inf)
   }
 
+  private def getToken(req: HttpServletRequest) = {
+    Option(req.getHeader("Authorization")) map {
+      _.stripPrefix("Bearer ")
```
rtitle, line 73: This could do a little more validation that it is indeed a bearer token. E.g. if we get garbage authorization headers it would be good to catch that before calling Sam.
```
private def getToken(req: HttpServletRequest): Option[String] = {
  Option(req.getHeader("Authorization")) flatMap { header =>
    val split = header.split(" ")
    if (split.length == 2 && split(0) == "Bearer") Option(split(1))
    else None
  }
}
```

src/main/scala/org/broadinstitute/dsde/workbench/avram/dataaccess/HttpSamDao.scala
```diff
@@ -0,0 +1,34 @@
...
+import io.circe.parser._
+import org.broadinstitute.dsde.workbench.avram.util.{ErrorResponse, RestClient}
+
+class HttpSamDao extends SamDao with RestClient {
+  private val log: Logger = Logger.getLogger(getClass.getName)
```
rtitle: It seems `RestClient` has a `Logger`, could you make it `protected` and use that one?

Also could consider mixing in a trait like `LazyLogging` (see https://github.com/lightbend/scala-logging).

src/main/scala/org/broadinstitute/dsde/workbench/avram/dataaccess/HttpSamDao.scala
rtitle: Yeah, definitely try to use `map`/`flatMap` instead of `match`. I think this could definitely be a for-comprehension:
```
val userInfoOrError = for {
  content <- response.body
  json <- parse(content)
  userInfo <- json.hcursor.get[UserInfo]("userInfo")
} yield userInfo

userInfoOrError.fold {
  error => // handle error
  result => result
}
```

I didn't compile but you get the idea..

src/main/scala/org/broadinstitute/dsde/workbench/avram/util/ErrorResponse.scala
rtitle: +1

src/main/scala/org/broadinstitute/dsde/workbench/avram/util/RestClient.scala
rtitle: Not entirely sure.. probably for our use case (and given GAE's restrictions) `HttpURLConnectionBackend` is appropriate.

rtitle:
Looks good in general, I like the approach using sttp, circe. 

Using `Either` is good too, but it'd be great to use the more idiomatic/functional utilities `map`, `flatMap`, `fold`, etc over `match`. Cats further extends `Either` by adding its set of typeclasses which might be useful someday. See: https://typelevel.org/cats/datatypes/either.html

It'd be good to add unit tests for the routes, maybe with a mock Sam (or stubbed `UserInfo`-extractor).

Lastly I tried to import avram into Intellij but I keep getting this:
```
[error] You need to set the 'APPENGINE_SDK_HOME' environment variable or the 'sdkHome' property in 'appengine.properties'
```
I installed the AppEngine SDK and I can build Avram on the command line using sbt. Any tips on how to get it working in Intellij?

src/main/scala/org/broadinstitute/dsde/workbench/avram/dataaccess/HttpSamDao.scala
rtitle: That is very cool

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
breilly2: Looks like this makes sure there aren't spaces in the token (which I guess are illegal in a token?) and that the token isn't `""`. I also wonder if there's a maximum length to the token that we can check for.

src/main/scala/org/broadinstitute/dsde/workbench/avram/dataaccess/HttpSamDao.scala
breilly2: Might be able to use `scala-logging`. Ultimately for App Engine, it has to boil down to `java.util.logging`. `getClass.getName` is going to return the name of the class that owns the logger so I think the logger needs to be in the immediate class (at least with `java.util.logging`).

src/main/scala/org/broadinstitute/dsde/workbench/avram/dataaccess/HttpSamDao.scala
breilly2: Yup. Just need to harmonize the error types, which is what I was illustrating by including the type annotations. They're all coming from different frameworks. sttp's `request.send()` gives a `String`, circe's `parse()` gives a `ParsingFailure`, and circe's `get()` gives a `DecodingFailure`. Circe's errors might be enough alike to map over (as long as they share the definition of `message()`), but I think it would be a problem to include sttp's `String` error in that for comprehension, right?

src/main/scala/org/broadinstitute/dsde/workbench/avram/util/RestClient.scala
breilly2: Well, as long as any threads are just used in the scope of the request, I think we're okay in GAE standard.

Though we might need to move from automatic to one of the other scaling types anyway. I'm finding more and more caveats with automatic, e.g. https://cloud.google.com/endpoints/docs/frameworks/known-issues#poor_performance_on_automatic_scaling_instances

rtitle: I don't really understand what "in the scope of the request" means w.r.t. threading. Does this mean request handler threads can submit tasks to a thread pool? Or does all work need to be synchronous to a request handler thread?

That doc you linked is slightly concerning.. :)

src/main/scala/org/broadinstitute/dsde/workbench/avram/dataaccess/HttpSamDao.scala
rtitle: Ohh right, the error types need to match. I wonder if we could define an `AvramException` type and automatically (implicitly?) convert the various cicre/sttp errors to the unified exception type.

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
rtitle: It also makes sure it's a `Bearer` token and not some other Auth type like `Basic`, etc.

My main point is to do some sanity checking of the Auth header before sending the token to Sam. Does `sttp` have any utilities for parsing common headers like `Authorization`?

rtitle: https://sttp.readthedocs.io/en/latest/requests/authentication.html

src/main/scala/org/broadinstitute/dsde/workbench/avram/util/RestClient.scala
breilly2: App Engine standard allows threads so we shouldn't be afraid of them. However, App Engine standard with automatic scaling does not support _background_ threads. (I don't think it explicitly fails but it also doesn't provide CPU time for them to run and, due to the automatic scaling, we can't rely on any background threads safely completing their tasks.) By "in the scope of the request" I mean while the main request thread is alive. During that time, we are absolutely allowed to have threads running. For example, if we had 2 independent web service calls to make, we could use something other than `HttpURLConnectionBackend` to run them in parallel and collect the results to build our response; we don't have to run them synchronously.

rtitle: Ah I see - thank you.

src/main/scala/org/broadinstitute/dsde/workbench/avram/dataaccess/HttpSamDao.scala
```diff
@@ -0,0 +1,34 @@
...
+  // TODO: read from config
+  val samUrl = "https://sam.dsde-dev.broadinstitute.org"
+
+  override def getUserStatus(token: String): Either[ErrorResponse, UserInfo] = {
+    val request = buildAuthenticatedGetRequest(samUrl, "/register/user/v1?userDetailsOnly=true", token)
```
dvoet: please use the v2 end point https://sam.dsde-dev.broadinstitute.org/#!/Users/getUserStatusInfo

src/main/scala/org/broadinstitute/dsde/workbench/avram/dataaccess/HttpSamDao.scala
breilly2: That sounds like a good idea and is the kind of feedback/advice I was hoping for. Thanks! I'm also absolutely thrilled that I understood what was going on well enough to recognize and explain it in a way that made sense! #fp_takeover_plot

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
dvoet: also the check on `Bearer` should be case insensitive

breilly2: @rtitle That API is for using sttp as a client. This code is getting the token from the `HttpRequestServlet` that cloud endpoints framework is providing to our endpoint function. The real question is does cloud endpoints framework support parsing the `Authorization` header. The answer is complicated and I'd have to re-read some docs to give you a full answer, but the short version is that cloud endpoints wants to do all the authn and give our function a `User` object or something like that. We want to instead implement authn as a call to sam and I don't think there's a way for us to insert that into cloud endpoint's authn flow.

src/main/scala/org/broadinstitute/dsde/workbench/avram/dataaccess/HttpSamDao.scala
breilly2: @dvoet The v2 endpoint does not have a `userDetailsOnly` parameter. Is that a problem or is the v2 endpoint inherently more efficient?

ansingh7115:
Question: will this PR include a cache for tokens so we don't have to hit sam each time?

breilly2:
I hadn't thought about using a cache. My initial thought was that we probably can't cache anything but the proxy used by other services has a cache; we should learn more about that. May want to not cache authz but maybe we can cache authn. Bernick may have some thoughts. I'll check with him.

src/main/scala/org/broadinstitute/dsde/workbench/avram/util/RestClient.scala
ahaessly: i'm concerned about the threading for when we need to support asynchronous calls like loading entities from a google bucket.

dvoet:
gae has built in memcache, you can cache the authn token for the lifetime of the token

breilly2:
@dvoet Thanks, that makes sense. When we get to authz, I'm thinking we should _not_ cache those results, right?

rtitle:
@breilly2 looks like there are some recent changes since I last looked at this, let me know when/if you want me to take another pass at this.

breilly2:
@rtitle This PR has become too large for my liking and it covers too many different concerns. I'm going to try breaking out a bunch of smaller PRs (the first of which is #8) while making sure I address the feedback here along the way.

breilly2:
This PR was split into a bunch of smaller PRs.

PR #8, opened 2018-09-21
Proposed code structure for locating services (no JIRA)

breilly2:
On my way home, I decided I was being lazy by including these changes in my other PR since they really have nothing to do with that one. Please review this one first, and soon if you can. It's small and I think it should be non-contentious so I'd like to merge it quickly.

I recommend starting with `Avram.scala` and `AvramConfig.scala` and saving `AvramRoutes.scala` for the end. The rest is up to you.

In all cases:


src/main/scala/org/broadinstitute/dsde/workbench/avram/Avram.scala
```diff
@@ -0,0 +1,43 @@
...
+  *
+  * The need for this is a consequence of Google Cloud Endpoints Framework controlling instance
+  * creation of the API implementation objects. However, introducing Guice
+  * (https://cloud.google.com/endpoints/docs/frameworks/java/using-guice) might give that control
+  * back to us and might be worth exploring.
+  */
+object Avram {
+  private val databaseFactory = new SlickDatabaseFactory(AvramConfig.dbcpDataSourceConfig)
+
+  val database = databaseFactory.database
```
rtitle, line 36: We may want to wrap the slick `database` in a class with utilities for using transactions, accessing data, etc. Right now Avram just runs raw sql queries but we probably want to lift that into a DB component layer.

For example Leo uses the cake pattern to break up database tables into different components, and provides a `DbReference` class for accessing the components within a transaction.

See: https://github.com/DataBiosphere/leonardo/blob/develop/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/DbReference.scala#L59

Usage:
```
dbRef.inTransaction { dataAccess =>
  dataAccess.clusterComponent.list()
}
```

I am wondering if it makes more sense to encapsulate `databaseFactory`, `database`, and `dbcpDataSource` in some sort of `DbReference` class, and have the `Avram` object just provide the `DbReference`.

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
```diff
@@ -45,15 +38,16 @@ class AvramRoutes {
   // TODO: move/merge this endpoint into a status API
   @ApiMethod(name = "dbPoolStats", httpMethod = "get", path = "dbPoolStats")
   def dbPoolStats: DbPoolStats = {
+    val dataSource = Avram.dbcpDataSource
     val result = for {
-      totalConnections <-  dataSource.database.run(
+      totalConnections <- database.run(
         sql"select count(*) from pg_stat_activity where pid <> pg_backend_pid() and usename = current_user".as[Int])
```
rtitle, line 44: Alluded to above, but would be nice to move SQL queries out of the routes, into a DB component layer.

src/main/scala/org/broadinstitute/dsde/workbench/avram/Avram.scala
breilly2: That sounds right to me. Anu is working on putting collections into the DB. I think we can apply the refactor you suggest here once we have something in the database to manage.

(I'll have to read up about the cake pattern again.)

src/main/scala/org/broadinstitute/dsde/workbench/avram/util/SlickDatabaseFactory.scala
```diff
@@ -0,0 +1,45 @@
...
+import slick.jdbc.PostgresProfile.api._
+
+/**
+  * Provides access to a slick Database object using the provided configuration.
+  *
+  * @param dataSourceConfig database and connection pool configuration
+  */
+class SlickDatabaseFactory(dataSourceConfig: DbcpDataSourceConfig) {
+
+  // See https://commons.apache.org/proper/commons-dbcp/configuration.html for configuration options and defaults
```
rtitle: Maybe lift this into a `makeDbcpDataSource` method? (similar to `makeHikariDataSource`)

breilly2: Good call. `makeHikariDataSource` was mostly a cheat to keep unused code in the commit while removing conceptual complexity for the reviewer.

rtitle (approved):
OK with me to merge this, with the caveat of restructuring the DB stuff later once collections are added. 

The `Avram` service registry seems ok to me. We could look at Guice for DI (I've used it with Java but never with scala). There are other more FP alternatives to DI like Reader monad which we could explore. :) But this should be fine for now.

breilly2:
I definitely want to chat about Reader. I still don't have an intuition for where to apply these things so it didn't jump out at me. If that works out, I suspect it will be much better than trying to use Guice. My feeling is that we're still dead in the water by not being in control of instantiation of the API endpoint instances but I'd be happy to be proven wrong. :)

src/main/scala/org/broadinstitute/dsde/workbench/avram/Avram.scala
ansingh7115: Yep, working on a PR to add Transactions and component layer!

rtitle:
Yeah I'd love to discuss `Reader`! It's not really a magic bullet -- it just lets you access your dependencies in a monadic way and "inject" them at the last possible moment (in our case, when the API response is committed). You still have to write the actual wiring for injection.

For example, I could see us defining a monad stack for the different "effects" we want:
- `Reader` for dependency injection
- `Future` (`IO`?) for asynchrony
- `Either` for error handling

and defining all Avram routes in terms of this stack. Then provide some kind of (implicit?) way of "unwrapping" the stack:
- injecting the dependency into `Reader` 
- awaiting the `Future`
- `fold`ing the `Either` to get a result or an error

This is a little off-topic for this PR... I still think the current approach is fine for now. The above needs some experimentation to see if it's feasible.

breilly2:
Yes, that makes sense. I'd love to pair on this for an hour or two to explore it sometime.

I experimented a little with Reader based on http://blog.originate.com/blog/2013/10/21/reader-monad-for-dependency-injection/. Ultimately, though, what we don't have is injection of any kind at the highest level entry point of the application. From the article:
```
object Application extends Application(UserRepositoryImpl)

class Application(userRepository: UserRepository) extends Controller with Users {
  ...
}
```
This illustrates things we don't have in a GAE/Cloud Endpoints world. Because Google's `EndpointServlet` instantiates the endpoint instance (equivalent of the `Controller` aspect of `Application`), we can't give it constructor parameters and provide concrete implementations of the dependencies (unless we do the Guice thing). We can probably do something like this at the level of the `Avram` object, but ultimately the endpoint instance needs to be able to discover it without any help.

Now that I think about it, maybe that's what you intended with your comment. I can play around some more.

rtitle:
I was also kind of hand-wavy. :) Sure would love to pair on this sometime (maybe next week?). Again I don't think it should block this PR.

ahaessly (approved):
👍

PR #9, opened 2018-09-22
GAWB-3754: Add separate liquibase deploys and Slick components for Collections and Entitites with JSONB support

ansingh7115:
This needs some improvements but the big stuff works

In all cases:


db/db-deploy.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
+# Right now this is 1 line because all we want to is update the db, but there might be other deploy steps
+mvn liquibase:update -e
```
ansingh7115: This deploy script is 1 line right now. In the future, we'd want a jenkins job that runs this. Right now, it can be run manually.

db/pom.xml
```diff
@@ -0,0 +1,36 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
```
ansingh7115, line 3: The liquibase deploy is currently using mvn to update. Could change it to use an sbt/liquibase plugin potentially or keep it like this to keep the dependencies separate? idk

project/Dependencies.scala
```diff
@@ -97,6 +99,8 @@ object Dependencies {
     scalaTest,
 
     slick,
+    slickPg,
+    slickPgSpray,
```
ansingh7115: I think we were saying we want to switch to Circe but Brian said he had some reservations. We can always change this to work with Circe once we decide.

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
```diff
@@ -27,33 +31,45 @@ class AvramRoutes {
...
-    Now(Await.result(Future(fetchTimestampFromDBWithSlick()), Duration.Inf))
+  //I'm going to delete this obvs but here's how we might use a Transaction
+  @ApiMethod(name = "addCollection", httpMethod = "get", path = "addCollection")
+  def addCollection: Collection = {
+    Await.result(
+      for {
+        _ <- dataReference.inTransaction { dataAccess => dataAccess.collectionQuery.save("testResource", "samResource1") }
+      } yield {
+        Collection("testResource", "samResource2")
+      }, Duration.Inf)
```
ansingh7115: This is going to be deleted, just wanted to show the transaction. I've tested adding entities with JSONB and it works.

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/AvramPostgresProfile.scala
```diff
@@ -0,0 +1,20 @@
+package org.broadinstitute.dsde.workbench.avram.db
+
+import slick.jdbc.PostgresProfile
+import com.github.tminglei.slickpg._
+import slick.basic.Capability
+import slick.jdbc.JdbcCapabilities
+
+trait AvramPostgresProfile extends PostgresProfile with PgJsonSupport with PgSearchSupport with PgSprayJsonSupport {
```
ansingh7115: I don't thing we need PgSearchSupport, I'll test it without it, that was a copy paste leftover

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/EntityComponent.scala
```diff
@@ -0,0 +1,50 @@
...
+    def uniqueKey = index("IDX_NAME_COLLECTION_UNIQUE", (name, collection), unique = true)
+
+    def * = (id, name, collection, entityBody) <> (EntityRecord.tupled, EntityRecord.unapply)
+  }
+
+  object entityQuery extends TableQuery(new EntityTable(_)) {
+
+    def save(name: String, collection: Long, entityBody: JsValue): DBIO[Int] = {
+      entityQuery += EntityRecord(0, name, collection, entityBody)
+    }
```
ansingh7115, line 65: we can use JsValue for the JSONB. The funny thing is that when I pass it up to Cloud Endpoints in a case class like so: `Entity(....@BeanProperty entityBody: JsValue)`, it doesn't know what to do with a JsValue so doing a get returns:

{
 "name": "entityName1",
 "collection": "collection1",
  "entityBody": { }
}

where the entityBody is empty. This doesn't happen if the JsValue is turned into a string, obvs, but it would be nice if Cloud Endpoints could handle a JsValue.

src/test/scala/org/broadinstitute/dsde/workbench/avram/db/CollectionComponentSpec.scala
```diff
@@ -0,0 +1,19 @@
...
+
+  "CollectionComponent" should "save" in isolatedDbTest {
+
+    val collection = Collection("collection1", "samResource1")
+    dbFutureValue { _.collectionQuery.save(collection.name, collection.samResource) }
+
+    val result = dbFutureValue { _.collectionQuery.getCollectionByName("collection1") }
+    result shouldEqual Some(collection)
+
+  }
```
ansingh7115, line 25: Have to add in more tests here and an EntityComponentSpec

db/liquibase/changesets/20180915_collection.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="avram" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="avram" author="ansingh" id="collection11">
+        <createTable tableName="COLLECTION">
+            <column name="id" type="BIGINT" autoIncrement="true">
+                <constraints primaryKey="true"/>
+            </column>
+            <column name="name" type="VARCHAR(254)">
```
rtitle: Nit: `VARCHAR(255)` is a common convention, I don't know why all our other stuff uses 254. :)

db/liquibase/changesets/20180915_collection.xml
```diff
@@ -0,0 +1,16 @@
...
+                <constraints nullable="false" unique="true"/>
+            </column>
+            <column name="samResource" type="VARCHAR(254)">
+                <constraints nullable="false"/>
+            </column>
+        </createTable>
```
rtitle, line 26: It would be good to add audit columns to this table. Something like:
```
createdBy varchar(255) not null
createdTimestamp timestamp not null
deletedBy varchar(255) not null
deletedTimestamp timestamp
```
And come to think of it collections are mutable right? So maybe we'd want `updatedBy` and `updatedTimestamp` as well.

db/liquibase/changesets/20180917_entity.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="avram" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="avram" author="ansingh" id="entity6">
```
rtitle: nit: why `entity6`?

db/liquibase/changesets/20180917_entity.xml
```diff
@@ -0,0 +1,22 @@
...
+        <createTable tableName="ENTITY">
+            <column name="id" type="BIGINT" autoIncrement="true">
+                <constraints primaryKey="true"/>
+            </column>
+            <column name="name" type="VARCHAR(254)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="collection" type="BIGINT">
```
rtitle: Nit: maybe name this `collectionId`

db/liquibase/changesets/20180915_collection.xml
```diff
@@ -0,0 +1,16 @@
...
+    </changeSet>
```
rtitle, line 30: Would a unique constraint on `name` make sense here?

db/liquibase/changesets/20180917_entity.xml
```diff
@@ -0,0 +1,22 @@
...
+                <constraints nullable="false"/>
+            </column>
+            <column name="entityBody" type="JSONB">
+                <constraints nullable="false"/>
+            </column>
```
rtitle, line 28: Same comment as above about audit columns.

db/liquibase/changesets/20180915_collection.xml
rtitle: Also potentially an index if we are going to be querying for collections by name.

Side question: what field will eventually be exposed to users: `name` or `id`?

db/liquibase/changesets/20180917_entity.xml
```diff
@@ -0,0 +1,22 @@
...
+        </createTable>
+        <addUniqueConstraint columnNames="name, collection" constraintName="IDX_ENTITY_COLLECTION_UNIQUE" tableName="ENTITY"/>
+        <addForeignKeyConstraint baseTableName="ENTITY" baseColumnNames="collection" constraintName="FK_COLLECTION"
```
rtitle: Same as above about also adding indexes on the fields this table will be queried by.

db/liquibase/liquibase.properties
```diff
@@ -0,0 +1,6 @@
+driver=org.postgresql.Driver
+classpath=postgresql-42.2.5.jar
+url=jdbc:postgresql://google/avram?ssl=false&socketFactoryArg=broad-avram-dev:us-central1:avram&socketFactory=com.google.cloud.sql.postgres.SocketFactory
```
rtitle: I imagine this will need to be templated somehow to deploy to multiple environments. (Or maybe we have separate ones per env.) Doesn't need to be for this PR, just thinking out loud.

db/pom.xml
rtitle: This works for me

project/Dependencies.scala
rtitle: Brian's other PRs use circe... we should definitely standardize on 1 json library, and circe >>> spray-json IMHO :)

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/AvramComponent.scala
```diff
@@ -0,0 +1,15 @@
...
+
+import slick.jdbc.JdbcProfile
+
+import scala.concurrent.ExecutionContext
+
+trait AvramComponent {
+  val profile: JdbcProfile
+  implicit val executionContext: ExecutionContext
+
+  protected final val dummyDate: Instant = Instant.ofEpochMilli(1000)
```
rtitle: I think this came from Leo to work around a MySQL bug? Can probably remove this if it's not needed here.

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/DbReference.scala
```diff
@@ -0,0 +1,45 @@
...
+  val ds = new BasicDataSource()
+  ds.setDriverClassName(dataSourceConfig.driverClassName)
+  ds.setUrl(dataSourceConfig.url)
+  ds.setUsername(dataSourceConfig.username)
+  ds.setPassword(dataSourceConfig.password)
+  ds.setMaxTotal(dataSourceConfig.maxTotal)
+
+  val database = Database.forDataSource(ds, Option(ds.getMaxTotal), AsyncExecutor("Avram Executor", dataSourceConfig.slickNumThreads, dataSourceConfig.slickQueueSize))
+
+  def inTransaction[T](f: (DataAccess) => DBIO[T], isolationLevel: TransactionIsolation = TransactionIsolation.RepeatableRead): Future[T] = {
```
rtitle: I think ReadCommitted is the default postgres isolation level. Maybe it should be the default here as well.

(Should think more about what type of transactions avram needs -- not sure at the moment).

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/EntityComponent.scala
rtitle: Yeah that's not ideal, not sure of the solution though..

rtitle (changes requested):
Nice work! See comments above

db/liquibase/liquibase.properties
breilly2: We'll also want to be able to run this against a local DB instance.

db/liquibase/changesets/20180915_collection.xml
ahaessly: the id will be used in the api. the name can be a user friendly name (i think it should be optional - we don't need it in the case of workspaces)

db/liquibase/changesets/20180915_collection.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="avram" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="avram" author="ansingh" id="collection11">
+        <createTable tableName="COLLECTION">
+            <column name="id" type="BIGINT" autoIncrement="true">
+                <constraints primaryKey="true"/>
```
ahaessly, line 6: i was thinking that the workspace could pass in its own id as the collection id. i was thinking that it would need to be a string, but now i'm remembering that there is some strange conversion between the uuid and how it is stored in the rawls database.

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/EntityComponent.scala
```diff
@@ -0,0 +1,50 @@
...
+
+trait EntityComponent extends AvramComponent  {
+  this: CollectionComponent =>
+
+  class EntityTable(tag: Tag) extends Table[EntityRecord](tag, "entity") {
+    def id =             column[Long]      ("id", O.PrimaryKey, O.AutoInc)
+    def name =           column[String]    ("name", O.Length(254))
+    def collection =     column[Long]      ("collection")
+    def entityBody =     column[JsValue]   ("entityBody", O.Length(254), O.SqlType("JSONB"))
+
```
ahaessly, line 36: wondering if you left out _type on purpose? there was some thinking that we shouldn't specify the type until we have validation in place.

db/pom.xml
breilly2: Ultimately I think it would be good to consolidate both the dependencies and required build tools but this is okay for now.

project/Dependencies.scala
breilly2: Ah, I need to clarify. My reservation isn't with circe itself but with automatic derivation. We can still choose semi-automatic derivation or go with fully manual.

db/liquibase/changesets/20180915_collection.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="avram" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="avram" author="ansingh" id="collection11">
+        <createTable tableName="COLLECTION">
+            <column name="id" type="BIGINT" autoIncrement="true">
```
breilly2: I think "autoincrement" is a thing in mysql so I looked into this a little. It looks like (though I haven't definitively confirmed) this translates to `SERIAL` in postgres which is basically a macro for a sequence and an integer column with a default value set from the sequence. I don't think there's anything to do here... it's just something you have to know about how liquibase supports postgres.

db/liquibase/changesets/20180915_collection.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="avram" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="avram" author="ansingh" id="collection11">
```
breilly2: This collection goes up to 11?

db/liquibase/changesets/20180915_collection.xml
breilly2: If we need an external identifier (alternate key), that should be a different column. We absolutely should have our own unique ID.

db/liquibase/changesets/20180915_collection.xml
breilly2: Also, I see no reason to limit this to a couple hundred characters. I think databases these days are pretty good at optimizing storage so there shouldn't be a penalty for increasing the width of `varchar` fields (though we should check this about postgres; most of my experience is with oracle).

I just did a little reading here: https://www.postgresql.org/docs/10/static/datatype-character.html and here: https://stackoverflow.com/a/4849030/244191. TL;DR: all character types perform reasonably similar. Max size of a column is about 1G. Also postgres adds an extension to `varchar` that allows you to not specify a length as well as a `text` type that is the same as an unbounded `varchar`.

I propose we use something like `varchar(1000)` to allow for a long name and protect us from excessive data storage requirements.

db/liquibase/changesets/20180917_entity.xml
breilly2: Or `collection_id`. I think snake_case is conventional for SQL databases (unless postgres has separate conventions).

breilly2: The comment about snake_case actually applies to everything in these changesets.

db/liquibase/changesets/20180915_collection.xml
breilly2: I think we don't need `logicalFilePath`: http://www.liquibase.org/documentation/databasechangelog.html

db/liquibase/changesets/20180915_collection.xml
```diff
@@ -0,0 +1,16 @@
...
+        <createTable tableName="COLLECTION">
```
breilly2: Identifiers in postgres are not case sensitive (unless they're quoted but we shouldn't be worrying about that). I propose that we always type identifiers in lowercase to avoid hurting our fingers from contortions while typing, but I'm open to other suggestions.

db/liquibase/changesets/20180915_collection.xml
breilly2: Same proposal for using lowercase when specifying DB column types.

db/liquibase/changesets/20180915_collection.xml
ansingh7115: Haha, no, I was just incrementing the number up as I was dropping/adding tables when I was testing this and then I never changed the ids back :)

db/liquibase/changesets/20180915_collection.xml
breilly2: Alternately, I bet we could just use `type="bigserial"` and let postgres do the work instead of liquibase. It depends on whether or not we want to make our liquibase changesets database-agnostic. Since we're planning to use postgres-specific features (for JSON), I think there's more value in embracing postgres and avoiding the indirection of the liquibase abstractions. Therefore, I propose we make it:
```
<column name="id" type="bigserial">
```

db/liquibase/changesets/20180915_collection.xml
```diff
@@ -0,0 +1,16 @@
...
+            <column name="id" type="BIGINT" autoIncrement="true">
+                <constraints primaryKey="true"/>
+            </column>
+            <column name="name" type="VARCHAR(254)">
+                <constraints nullable="false" unique="true"/>
```
breilly2, line 9: The tech doc shows collection name as optional (and therefore also does not need to be unique).

db/liquibase/changesets/20180917_entity.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="avram" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="avram" author="ansingh" id="entity6">
+        <createTable tableName="ENTITY">
+            <column name="id" type="BIGINT" autoIncrement="true">
+                <constraints primaryKey="true"/>
+            </column>
+            <column name="name" type="VARCHAR(254)">
```
breilly2: I might be forgetting previous conversations but I wouldn't expect there to be a required name property on an entity, in part because this starts to be opinionated about the entity schema (all entities _must_ have a "name" property).

db/liquibase/changesets/20180915_collection.xml
ansingh7115: Agreed for createdBy and createdTimestamp, but I'm not sure if we're persisting deleted collections in the db or if we're just deleting them from the db? I thought we were starting out just clean deleting them and not persisting and then adding that if people expressed a desire for it. 

Also, I think updatedby/updatedTimestamp (and deleted, if we add it) would need to be nullable?

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/DbReference.scala
```diff
@@ -0,0 +1,45 @@
...
+    import dataAccess.profile.api._
+    database.run(f(dataAccess).transactionally.withTransactionIsolation(isolationLevel))
+  }
+}
+
+class DataAccess(val profile: JdbcProfile)(implicit val executionContext: ExecutionContext) extends AllComponents {
+
+  def truncateAll(): DBIO[Int] = {
```
breilly2, line 36: This should only be called from tests. It would be great to find a way to have this function only live in src/test.

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/AvramPostgresProfile.scala
breilly2: I wonder if we can do without `PgSprayJsonSupport`... maybe replacing it with something else... in the interest of avoiding use of spray.

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/AvramPostgresProfile.scala
```diff
@@ -0,0 +1,20 @@
...
+  override def pgjson = "jsonb "
```
breilly2: Nit: extra trailing space

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/AvramComponent.scala
ansingh7115: Actually, I think we'll need the same bug fix if we're using created/updated/destroyed dates for entities and collections :)

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/DbReference.scala
breilly2: Yes, `ReadCommitted` is the default for postgres. I think we should stick with that if we can. It's a little extra work to consider the consequences in application code, but I think it's worth it to help limit transaction/lock contention.

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/EntityComponent.scala
```diff
@@ -0,0 +1,50 @@
...
+case class EntityRecord(id: Long, name: String, collection: Long, entityBody: JsValue)
+
+trait EntityComponent extends AvramComponent  {
+  this: CollectionComponent =>
+
+  class EntityTable(tag: Tag) extends Table[EntityRecord](tag, "entity") {
+    def id =             column[Long]      ("id", O.PrimaryKey, O.AutoInc)
+    def name =           column[String]    ("name", O.Length(254))
+    def collection =     column[Long]      ("collection")
+    def entityBody =     column[JsValue]   ("entityBody", O.Length(254), O.SqlType("JSONB"))
```
breilly2: Based on https://github.com/tminglei/slick-pg#configurable-typemappers, it looks like we could use circe instead of spray-json, which would change this `spray.json.JsValue` into `io.circe.Json`.

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/EntityComponent.scala
breilly2: I'm thinking we might want to serialize our responses into JSON strings ourselves (using circe) instead of trying to get Endpoints Framework to do it for us. That might help with this... especially if we switch from spray-json's `JsValue` to circe's `Json`. I just haven't had a chance to experiment with this yet.

src/test/scala/org/broadinstitute/dsde/workbench/avram/db/DbSingleton.scala
```diff
@@ -0,0 +1,16 @@
+package org.broadinstitute.dsde.workbench.avram.db
+
+import com.typesafe.config.ConfigFactory
+import net.ceedubs.ficus.Ficus._
+import org.broadinstitute.dsde.workbench.avram.TestExecutionContext
+import org.broadinstitute.dsde.workbench.avram.config.DbcpDataSourceConfig
+
+
+// initialize database tables and connection pool only once
+object DbSingleton {
```
breilly2, line 11: I'm not sure why we need this. May be best to chat in person. My feeling is that, if we need this, we're doing something wrong in setting up our app/config.

db/liquibase/changesets/20180915_collection.xml
ansingh7115: I don't believe it should be optional - every collection should have a name that's unique - we could discuss this in person!

db/liquibase/liquibase.properties
```diff
@@ -0,0 +1,8 @@
+driver=org.postgresql.Driver
+classpath=postgresql-42.2.5.jar
+url=jdbc:postgresql://127.0.0.1:5432/testdb
+username=avram
+password=test
+changeLogFile=changelog.xml
```
ansingh7115, line 6: This currently has the test db url, username and password - we can template-ize this in a future PR

docker/run-postgres.sh
```diff
@@ -9,6 +9,12 @@ start() {
     # start up PostgreSQL
     echo "starting up postgres container..."
     docker run --name $CONTAINER -e POSTGRES_USER=avram -e POSTGRES_PASSWORD=test -e POSTGRES_DB=testdb -d -p 5432:5432 circleci/postgres:$POSTGRES_VERSION-alpine-ram
+    echo "sleeping for 5 seconds"
+    sleep 5s
+
+    echo "running liquibase"
+    cd db
+    ./db-deploy.sh
 }
```
ansingh7115, line 18: I added a sleep in this block so it waits for the db to be up before running liquibase

src/main/scala/org/broadinstitute/dsde/workbench/avram/Avram.scala
```diff
@@ -1,12 +1,17 @@
...
+import org.broadinstitute.dsde.workbench.avram.config.{AvramConfig, DbcpDataSourceConfig}
 import org.broadinstitute.dsde.workbench.avram.dataaccess.{HttpSamDao, SamDao}
-import org.broadinstitute.dsde.workbench.avram.util.SlickDatabaseFactory
+import org.broadinstitute.dsde.workbench.avram.db.DbReference
+
+import scala.concurrent.ExecutionContext
 
 /**
-  * Object providing access to all services. This merges configuration and service code to provide
+  * Class providing access to all services. This merges configuration and service code to provide
```
ansingh7115: Db needs the execution context passed in - so I changed this from an object to a class. Unsure if that's the exact right approach, interested to hear other opinions.

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
```diff
@@ -47,26 +53,15 @@ class AvramRoutes extends BaseEndpoint {
...
-  // TODO: remove this endpoint when we have more meaningful ways to test database queries
-  @ApiMethod(name = "now", httpMethod = "get", path = "now")
-  def now: Now = {
-    // Explicitly use a Future to make sure the implicit ExecutionContext is being used
-    Now(Await.result(Future(fetchTimestampFromDBWithSlick()), Duration.Inf))
+  @ApiMethod(name = "status", httpMethod = "get", path = "status")
+  def status: Status = {
+    //TODO: make this use HealthMoniter
+    inTransaction(_.sqlDBStatus()).map(_ => Status("ok")).head
   }
```
ansingh7115, line 44: Here's a sort-of status endpoint. We want this to use HealthMoniter eventually (or write something similar) but I'm not doing that this PR.

db/liquibase/changesets/20180917_entity.xml
ansingh7115: But then if I'm a user with a whole bunch of entities and I just ran an analysis and now I want to write back to some of my entities, how do I (or the service that's doing this) tell avram which entities to write to?

db/liquibase/changesets/20180915_collection.xml
breilly2: Updated can be non-null and be initially populated with the same values as created. Avoids having to conditionally look at the created fields if you want to know the last write.

src/main/scala/org/broadinstitute/dsde/workbench/avram/Avram.scala
breilly2: Let's explore this further. Changing this to a class means that every instance of a class that extends `BaseEndpoint` will get its own set of these objects, including the database connection pool.

project/Dependencies.scala
```diff
@@ -64,9 +64,11 @@ object Dependencies {
 
   val sam: ModuleID = "org.broadinstitute.dsde.sam-client" %% "sam" % samV
 
-  val slick: ModuleID =     "com.typesafe.slick" %% "slick"                 % slickV
+  val slickPg: ModuleID = "com.github.tminglei" %% "slick-pg" % "0.16.3"
+  val slickPgCirce: ModuleID = "com.github.tminglei" %% "slick-pg_circe-json" % "0.16.3"
+  val slick: ModuleID =     "com.typesafe.slick" %% "slick" % slickV
   val dbcp2: ModuleID = "org.apache.commons" % "commons-dbcp2" % dbcpV
-  val liquibase: ModuleID = "org.liquibase"       % "liquibase-core"        % "3.5.3"
+  val liquibase: ModuleID = "org.liquibase"  % "liquibase-core" % "3.5.3"
```
breilly2: It looks like we don't currently need a liquibase dependency.

src/main/scala/org/broadinstitute/dsde/workbench/avram/Avram.scala
breilly2: I created a PR (#16) that refactors this so it can be an object again. Let me know what you think.

src/main/scala/org/broadinstitute/dsde/workbench/avram/Avram.scala
```diff
@@ -32,14 +35,17 @@ import org.broadinstitute.dsde.workbench.avram.util.SlickDatabaseFactory
   * back to us and might be worth exploring.
   */
 object Avram {
-  private val databaseFactory = new SlickDatabaseFactory(AvramConfig.dbcpDataSourceConfig)
+  private val configFactory = ConfigFactory.parseResources("app.conf").withFallback(ConfigFactory.load())
+  private val dbcpDataSourceConfig = configFactory.as[DbcpDataSourceConfig]("dbcpDataSource")
+
+  val database =  DbReference(dbcpDataSourceConfig)
```
breilly2: This can be:
```
  val database = DbReference(AvramConfig.dbcpDataSourceConfig)
```
Then you don't need the two lines above to load the configs.

Also a nit: extra space after `=`.

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/AvramComponent.scala
```diff
@@ -0,0 +1,26 @@
...
+
+import java.sql.Timestamp
+import java.time.Instant
+
+import slick.jdbc.JdbcProfile
+
+trait AvramComponent {
+  val profile: JdbcProfile
+
+  protected final val dummyDate: Instant = Instant.ofEpochMilli(1000)
```
breilly2: Oh, I see. It's a magic value because nulls aren't considered equal. From https://www.postgresql.org/docs/9.0/static/indexes-unique.html:

> Null values are not considered equal. A multicolumn unique index will only reject cases where all indexed columns are equal in multiple rows.

So we only need this if we are marking as deleted instead of actually deleting _and_ we have some sort of name that we want to be unique _and_ we want users to be able to reuse names of previously "deleted" rows.

I have further comments about this if we keep it, but let's first have the discussion about whether or not to have a unique name requirement.

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/CollectionComponent.scala
```diff
@@ -0,0 +1,65 @@
...
+                            createdBy: String,
+                            samResource: String,
+                            createdTimestamp: Timestamp,
+                            updatedBy: String,
+                            updatedTimestamp: Timestamp)
+
+trait CollectionComponent extends AvramComponent {
+
+  class CollectionTable(tag: Tag) extends Table[CollectionRecord](tag, "collection") {
+    def id =                     column[Long]            ("id",                  O.PrimaryKey, O.AutoInc)
```
breilly2, line 23: Do we have to include the column options like `O.PrimaryKey`? According to the slick docs (http://slick.lightbend.com/doc/3.2.3/schemas.html), it looks like they're only used when creating DDL statements... a feature of slick we're not using because we're using liquibase instead. I think we can safely omit all of the slick column options.

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
```diff
@@ -1,21 +1,30 @@
...
-
 import scala.beans.BeanProperty
-import scala.concurrent.ExecutionContext.Implicits.global
-import scala.concurrent.duration.Duration
-import scala.concurrent.{Await, Future}
+
 
 case class Pong()
-case class Now(@BeanProperty message: String)
+case class Collection(@BeanProperty name: String,
```
rtitle: I am wondering if these case classes should be moved out of the routes and into a `model` package.

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/CollectionComponent.scala
```diff
@@ -0,0 +1,65 @@
...
+
+    def * = (id, name, samResource, createdBy, createdTimestamp, updatedBy, updatedTimestamp) <> (CollectionRecord.tupled, CollectionRecord.unapply)
+  }
+
+  object collectionQuery extends TableQuery(new CollectionTable(_)) {
+
+    def save(name: String, samResource: String, createdBy: String): DBIO[Int] = {
+      val now = Timestamp.from(Instant.now)
+      //currently the updatedBy and updatedTimestamp are the same as created at save time
+      collectionQuery += CollectionRecord(0, name, createdBy, samResource, now, createdBy, now)
```
breilly2: Not fond of the magic `0` and having to know that it's ignored. (I wonder if it's still ignored if we don't specify `O.AutoInc` on the column?) http://slick.lightbend.com/doc/3.2.3/queries.html looks to propose making the ID column an `Option` which makes this a little better (`None` instead of `0`). It also makes the custom * projection useful because you can use `id.?` so only the case class says `Option`, not the `Table`.

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
rtitle: HealthMonitor uses background threads (and akka actors) so I'm not sure it'd work as is in GAE standard. We should think about whether we want asynchronous status checks in Avram, and how to implement them ([task queues](https://cloud.google.com/appengine/docs/standard/java/building-app/taskqueues) might be an option?). This should probably be a backlog task -- this current solution is totally fine for now.

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/BaseEndpoint.scala
```diff
@@ -38,4 +48,10 @@ abstract class BaseEndpoint {
     } yield tokenMatch.group(1)
     token.toRight(ErrorResponse(401, "Missing token"))
   }
+
+  def inTransaction[T](f: (DataAccess) => DBIO[T]): T = {
+    Await.result(database.inTransaction(f) , Duration.Inf)
```
rtitle: I'd vote for setting some time limit instead of `Inf`. If the DB is unhealthy we want to abort the query and not wait forever. GAE's request timeout is 30s anyway (I think?) so a timeout around there might be reasonable.

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/AvramComponent.scala
rtitle: Hm - we don't currently have `deletedAt` columns, and the `createdAt` and `updatedAt` are not-null. So this isn't currently needed. Maybe remove this for now, and revisit adding `deletedAt` columns when we actually add the `DELETE` endpoints?

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
breilly2: Yes, I'd like to stay away from Akka.

By the way, threads that run in the background are perfectly fine in GAE. When they say things like "background threads are not allowed", they're talking about the deprecated GAE Background Threads API.

In an email from Ben Menasha (from Google), he said,
> The term "Background Threads" refers to app engine background threads which are currently deprecated : https://cloud.google.com/appengine/docs/standard/java/runtime#background_threads_1  and not to the creation of a new root threadgroup that outlives the request thread.

This info came out of my prodding about DB connection pools.

rtitle: Ah ok - so maybe a threaded (non-Akka) HealthMonitor implementation would work.

rtitle (approved):
This looks great!

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/CollectionComponent.scala
ansingh7115: I kind of like having column options, keys and indexes here so I can see the structure of the table - but I don't really have a strong opinion either way. @ahaessly @rtitle what do you think?

breilly2: Assuming they turn out to be completely non-functional, my strongest argument against having them is that one might think they can change behavior by changing the flags here.

rtitle: I didn't realize they were just used for DDL updates (but that makes sense). I kind of like having them -- they are useful as documentation and I worry about some undocumented slick behavior if we omit them. Maybe leave them but include a comment next to the record definitions that we use liquibase for DDL updates instead of slick?

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
breilly2: I wonder if they can be merged with the `*Record` case classes. I have mixed feelings about that because they are technically different things, but they are _so_ similar that the strict separation might be less valuable than the reduction of code.

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/CollectionComponent.scala
breilly2: If we want documentation, we can add documentation comments. I still feel that code with zero effect is distracting and misleading.

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/CollectionComponent.scala
```diff
@@ -0,0 +1,65 @@
...
+
+import org.broadinstitute.dsde.workbench.avram.api.Collection
+import AvramPostgresProfile.api._
+
+import scala.concurrent.ExecutionContext
+
+case class CollectionRecord(id: Long,
+                            name: String,
+                            createdBy: String,
+                            samResource: String,
```
breilly2, line 14: I was going to suppress this nit but it's actually a bug. The order of the fields here is different than the order of the columns in `CollectionTable`. The * projection uses the order from `CollectionTable` which means that the `createdBy` and `samResource` values are being swapped... and they're both `String` so there's no compilation error... and the technique of defining the * projection uses `CollectionRecord.tupled` so there's no way to avoid the problem by using named parameters.

Short term, we need to make sure the fields are consistently listed in the same order. Long term, I'd like to figure out how to craft the code to make these kinds of errors impossible (or at least more obvious).

rtitle: That's a good catch - yeah the order needs to be the same as in the `*` projection.

It's possible to define the projection not using `.tupled` so it's not dependent on the order. (It's more boilerplate though.)

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/CollectionComponent.scala
```diff
@@ -0,0 +1,65 @@
...
+                            createdTimestamp: Timestamp,
```
breilly2, line 16: I'd prefer to keep `java.sql` classes out of our domain model. Also, `java.sql.Timestamp` has issues (e.g. it's a `java.util.Date` but doesn't conform to the equals/hashCode contract... and appears to use precision if used as the key of a hash map).

Maybe we can convert it to a `java.time.Instant`?

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/DbReference.scala
```diff
@@ -0,0 +1,54 @@
...
+    ds.setUsername(dataSourceConfig.username)
+    ds.setPassword(dataSourceConfig.password)
+    ds.setMaxTotal(dataSourceConfig.maxTotal)
+    ds
+  }
+  val dbcpDataSource: BasicDataSource = makeDbcpDataSource
+
+  val database = Database.forDataSource(dbcpDataSource, Option(dbcpDataSource.getMaxTotal), AsyncExecutor("Avram Executor", dataSourceConfig.slickNumThreads, dataSourceConfig.slickQueueSize))
+
+  def inTransaction[T](f: (DataAccess) => DBIO[T], isolationLevel: TransactionIsolation = TransactionIsolation.ReadCommitted): Future[T] = {
```
breilly2: If `inTransaction` is the intended way to access the database, `database` above should be a private val. `PostgreSQLSpec` will need to be updated to use `inTransaction`.

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/BaseEndpoint.scala
```diff
@@ -1,11 +1,21 @@
...
 import org.broadinstitute.dsde.workbench.avram.util.ErrorResponse
+import slick.dbio.DBIO
+
+import scala.concurrent.Await
+import scala.concurrent.duration.Duration
 
 abstract class BaseEndpoint {
+
+  private val log = Logger.getLogger(getClass.getName)
+  val database = Avram.database
```
breilly2: The way that `AvramRoutes.dbPoolStats` uses this `database` property is definitely non-standard. Endpoints should be using the provided `inTransaction` function. Therefore, this `database` property should be private and `AvramRoutes.dbPoolStats` should be forced to access `Avram.database` directly.

This probably seems a bit picky but I think it's worth doing to establish the correct patterns now.

src/main/scala/org/broadinstitute/dsde/workbench/avram/Avram.scala
```diff
@@ -32,14 +35,15 @@ import org.broadinstitute.dsde.workbench.avram.util.SlickDatabaseFactory
...
   val samDao: SamDao = new HttpSamDao(AvramConfig.sam.baseUrl)
 
   /**
     * DBCP data source provided only for introspection into the database pool statistics. If you're
     * not working with database monitoring, don't use this!
     */
-  val dbcpDataSource: BasicDataSource = databaseFactory.dbcpDataSource
-}
\ No newline at end of file
+  val dbcpDataSource: BasicDataSource = database.dbcpDataSource
```
breilly2: Let's get rid of this since `AvramRoutes.dbPoolStats` can (and does) access it using `Avram.database.dbcpDataSource`. I don't know why I added this here.

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/CollectionComponent.scala
```diff
@@ -0,0 +1,65 @@
...
+  class CollectionTable(tag: Tag) extends Table[CollectionRecord](tag, "collection") {
+    def id =                     column[Long]            ("id",                  O.PrimaryKey, O.AutoInc)
+    def name =                   column[String]          ("name",                O.Unique, O.Length(1000))
+    def samResource =            column[String]          ("sam_resource",        O.Length(1000))
+    def createdBy =              column[String]          ("created_by",          O.Length(1000))
+    def createdTimestamp =       column[Timestamp]       ("created_timestamp",   O.SqlType("TIMESTAMP(6)"))
+    def updatedBy =              column[String]          ("updated_by",          O.Length(1000))
+    def updatedTimestamp =       column[Timestamp]       ("updated_timestamp",   O.SqlType("TIMESTAMP(6)"))
+
+    def * = (id, name, samResource, createdBy, createdTimestamp, updatedBy, updatedTimestamp) <> (CollectionRecord.tupled, CollectionRecord.unapply)
```
breilly2: Here's where I'm thinking that we could convert `java.sql.Timestamp`s to `java.time.Instant`s. One thing I'm not sure of is how (in slick) to write the conversions in both directions.

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/CollectionComponent.scala
breilly2: I realize that I misunderstood a little about the * projection. It is currently being used to convert between the tuple of values and the `CollectionRecord` case class. The point though is that it doesn't have to be a 1-1 mapping. We are allowed to do other things here if it helps bridge the object-relational gap.

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
ahaessly: wow! that really clears up a misinterpretation. i'll post around so others thinking about GAE standard know this is not a limitation.

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/CollectionComponent.scala
ansingh7115: Good catch!

src/test/scala/org/broadinstitute/dsde/workbench/avram/db/TestComponent.scala
```diff
@@ -0,0 +1,36 @@
...
+
+import scala.concurrent.ExecutionContext
+
+trait TestComponent extends Matchers with ScalaFutures with AvramComponent {
+  override val profile: JdbcProfile = DbSingleton.ref.dataAccess.profile
+  implicit val executionContext: ExecutionContext = TestExecutionContext.testExecutionContext
+  implicit override val patienceConfig = PatienceConfig(timeout = scaled(Span(10, Seconds)))
+  val defaultServiceAccountKeyId = ServiceAccountKeyId("123")
+
+  def dbFutureValue[T](f: DataAccess => DBIO[T]): T = DbSingleton.ref.inTransaction(f).futureValue
```
breilly2, line 21: It would be nice to have something that gives back the `Future` to allow full use of the `FutureConcept` functions. Something like:
```
def db[T](f: DataAccess => DBIO[T]): Future[T] = DbSingleton.ref.inTransaction(f)
```
This would allow test code like (adapted from `CollectionComponentSpec`):
```
    val futureResult = db(_.collectionQuery.getCollectionByName(CommonTestData.collectionName).map(_.get))
    whenReady(futureResult) { collection =>
      collection.name shouldEqual CommonTestData.collectionName
      collection.samResource shouldEqual CommonTestData.samResource
      collection.createdBy shouldEqual CommonTestData.user1
    }
```
Could also have a `dbRun` or something that uses `futureValue` to execute the thing but throw away the result (for cases where we only care about the effect, like for `truncateAll()`). Just gives us a little more flexibility in our test code.

src/test/scala/org/broadinstitute/dsde/workbench/avram/CommonTestData.scala
```diff
@@ -25,8 +32,15 @@ object CommonTestData {
     slickQueueSize = 1000
   )
 
-  val localDatabase = new SlickDatabaseFactory(localDataSourceConfig).database
+  val configFactory = ConfigFactory.parseResources("app.conf").withFallback(ConfigFactory.load())
+  private val dbcpDataSourceConfig = configFactory.as[DbcpDataSourceConfig]("dbcpDataSource")
+  val dataSource = new DbReference(dbcpDataSourceConfig)
+
+  val collectionName = "collection1"
```
breilly2, line 39: This can be fixed later, but I'll still point out that I think these common values aren't adding any value as they currently are... and I'm generally not in favor of shared test data anyway (though sometimes it's just too much work to create a full object graph in the test case). A case where these might be more useful is if they were paired with some correlated data, such as the same values being in a stubbed JSON response.

src/test/scala/org/broadinstitute/dsde/workbench/avram/CommonTestData.scala
```diff
@@ -25,8 +32,15 @@ object CommonTestData {
     slickQueueSize = 1000
   )
 
-  val localDatabase = new SlickDatabaseFactory(localDataSourceConfig).database
```
breilly2, line 28: I liked having "local" in the name to communicate that it's talking to a local DB as opposed to Cloud SQL. I think tests should be decisive about what environment they expect to run in and the naming clue helps readers more easily understand the test.

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/DbReference.scala
```diff
@@ -0,0 +1,54 @@
...
+    database.run(f(dataAccess).transactionally.withTransactionIsolation(isolationLevel))
```
breilly2: I'd rather we leave out setting the transaction isolation level, even if we're defaulting to the database default, until we find we need to change it (if ever).

db/liquibase/changesets/20180915_collection.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="avram" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet author="ansingh" id="add_collection_table">
+        <createTable tableName="collection">
+            <column name="id" type="bigserial" autoIncrement="true">
```
breilly2: Does `autoIncrement="true"` have any effect along with `type="bigserial"`? I suspect `autoIncrement="true"` is redundant. Did you try leaving it out?

db/liquibase/changesets/20180915_collection.xml
```diff
@@ -0,0 +1,31 @@
...
+                <constraints primaryKey="true"/>
+            </column>
+            <column name="name" type="varchar(1000)">
+                <constraints nullable="false" unique="true"/>
+            </column>
+            <column name="sam_resource" type="varchar(255)">
```
breilly2: I don't know what sam resource identifiers look like, but maybe we should make this wider (maybe 1000) just to be safe.

db/liquibase/changesets/20180915_collection.xml
```diff
@@ -0,0 +1,31 @@
...
+            </column>
+            <column name="updated_by" type="varchar(1000)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="updated_timestamp" type="timestamp">
+                <constraints nullable="false"/>
+            </column>
+        </createTable>
+        <createIndex tableName="collection" indexName="IDX_collection_name">
+            <column name="name" type="varchar(1000)"/>
```
breilly2: Does type need to be repeated when creating an index?

breilly2 (changes requested):
So... I'm sorry... I know I've given a lot of feedback. I didn't do a good job of filtering or indicating what I think should be fixed vs what can be a future improvement. If you're unsure, just ask (over slack if you want a quick-ish response).

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/CollectionComponent.scala
ansingh7115: I would like to resolve this in a future PR and not here.

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
ansingh7115: I would keep the Record classes separate for now - record classes have `id`s, which response classes don't. And in the future we may want to keep track of other fields in our record classes and not our response classes.

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/CollectionComponent.scala
ansingh7115: I would like to keep Record and Response classes separate so we can keep it like this

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/AllComponents.scala
```diff
@@ -0,0 +1,4 @@
+package org.broadinstitute.dsde.workbench.avram.db
+
+// a trait combining all of the individual LeoComponent traits
```
ahaessly: change Leo to Avram

db/liquibase/changesets/20180917_entity.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="avram" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="avram" author="ansingh" id="add_entity_table">
+        <createTable tableName="entity">
+            <column name="id" type="bigserial" autoIncrement="true">
+                <constraints primaryKey="true"/>
+            </column>
+            <column name="name" type="varchar(1000)">
+                <constraints nullable="false"/>
```
ahaessly: we should not require an entity name. it should be up to the person defining the model what fields are useful

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/EntityComponent.scala
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    def getEntityByName(name: String, collectionId: Long)(implicit executionContext: ExecutionContext): DBIO[Option[Entity]] = {
+      entityQuery
+        .filter { _.name === name}
+        .filter { _.collectionId === collectionId }
+        .result map { recs: Seq[EntityRecord] =>
+        unmarshalEntities(recs).headOption
+      }
+    }
```
ahaessly, line 59: instead of getEntityByName there will most likely be a getEntityByField where the user supplies the field name and value. this will be used when we reference an entity by $ref.

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/CollectionComponent.scala
ansingh7115: I think using a 0 for now is fine, we can talk about it as a future improvement to the code, but I don't want to explore it for this PR.

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/EntityComponent.scala
ansingh7115: Yeah, definitely - we can add the getEntityByField method when we implement the GET - I just wanted to write some bare bones stuff to test the db

src/test/scala/org/broadinstitute/dsde/workbench/avram/db/TestComponent.scala
ansingh7115: I don't want to add this in this PR

db/liquibase/changesets/20180915_collection.xml
ansingh7115: Either that or we'd need an external id column

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/CollectionComponent.scala
breilly2: This is about converting the row tuple from slick to the `*Record` instance, not `*Record` vs `*Response`. I want to insulate business logic from strangeness with `java.sql.Timestamp`, and from `java.sql` in general.

I can be okay with not addressing this in this PR as long as we consider this (and possibly other points) for a pre-prod TODO list.

breilly2: Oh, it looks like you did make this change. Never mind.

db/liquibase/changesets/20180915_collection.xml
ansingh7115: Changed to external id

db/liquibase/changesets/20180917_entity.xml
```diff
@@ -1,12 +1,14 @@
...
         <createTable tableName="entity">
             <column name="id" type="bigserial">
                 <constraints primaryKey="true"/>
             </column>
-            <column name="name" type="varchar(1000)"/>
-            <column name="collection_id" type="bigserial">
+            <column name="external_id" type="UUID">
+                <constraints nullable="false" unique="true"/>
+            </column>
+            <column name="external_collection_id" type="UUID">
```
breilly2, line 11: We might want to use `collection_id` as the foreign key. I don't have a solid reason for it other than my understanding of data modeling practices. `id` is the primary key while `external_id` is an alternate key. However, using the primary key means having a join on `collection` when selecting a collection's entities by `external_collection_id`.

I'm okay with keeping it like this to get this PR merged and thinking on it more before going to production.

db/liquibase/changesets/20180915_collection.xml
```diff
@@ -24,8 +24,8 @@
                 <constraints nullable="false"/>
             </column>
         </createTable>
-        <createIndex tableName="collection" indexName="IDX_collection_name">
-            <column name="name"/>
+        <createIndex tableName="collection" indexName="IDX_collection_external_id">
```
breilly2, line 27: Postgres creates an index automatically for unique constraints. I'm guessing that saying `unique="true"` on the column will do that but I haven't checked. We may find that we won't need this index, but that is easy enough to drop (and shouldn't have negative impacts on table storage... but who knows). Anyway, again, good to merge as-is.

breilly2 (approved):
👍 Thanks for responding to the feedback and making changes.

PR #10, opened 2018-09-25
Small refactor of common test data (no JIRA)

breilly2:
Another small incremental improvement.

In all cases:

### DIFF
--- a/src/test/scala/org/broadinstitute/dsde/workbench/avram/CommonTestData.scala
+++ b/src/test/scala/org/broadinstitute/dsde/workbench/avram/CommonTestData.scala
@@ -1,21 +1,32 @@
 package org.broadinstitute.dsde.workbench.avram
 
-import com.typesafe.config.ConfigFactory
 import org.broadinstitute.dsde.workbench.avram.config.DbcpDataSourceConfig
 import org.broadinstitute.dsde.workbench.avram.util.SlickDatabaseFactory
-import net.ceedubs.ficus.Ficus._
-import net.ceedubs.ficus.readers.ArbitraryTypeReader._
 
-// values common to multiple tests, to reduce boilerplate
+/**
+  * Values common to multiple tests, to reduce boilerplate.
+  */
+object CommonTestData {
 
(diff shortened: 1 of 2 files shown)


src/test/scala/org/broadinstitute/dsde/workbench/avram/CommonTestData.scala
```diff
@@ -1,21 +1,32 @@
...
-
-trait CommonTestData {
-
-  val configFactory = ConfigFactory.parseResources("app.conf").withFallback(ConfigFactory.load())
-  private val dbcpDataSourceConfig = configFactory.as[DbcpDataSourceConfig]("dbcpDataSource")
-  val dataSource = new SlickDatabaseFactory(dbcpDataSourceConfig)
+/**
+  * Values common to multiple tests, to reduce boilerplate.
+  */
+object CommonTestData {
```
breilly2, line 9: Making this an object makes is more like the `Avram` object, meaning the creation patterns and instance lifecycle are more similar. Specifically, if this were a trait, I'm pretty sure each instance of each test class extending it would get its own DB connection pool!

src/test/scala/org/broadinstitute/dsde/workbench/avram/integration/PostgreSQLSpec.scala
```diff
@@ -7,16 +7,15 @@ import slick.jdbc.PostgresProfile.api._
...
+  val database = CommonTestData.localDatabase
 
   "should fetch" - {
     "using slick" in {
-      val db = dataSource.database
-      try {
-        val action: DBIO[Seq[String]] = sql"select now()".as[String]
-        val value: Seq[String] = Await.result(db.run(action), Duration.Inf)
-        value foreach { v => println(s"from slick: $v") }
-      } finally db.close
```
breilly2, line 19: This is only a test. The connection will close when the process exits. No need for the extra verbosity here.

rtitle (approved):

ahaessly (approved):
👍

PR #11, opened 2018-09-25
GAWB-3765: Added SamDao

breilly2:
Note: not all of GAWB-3765. Just adding the SamDao.

In all cases:

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -79,7 +79,7 @@ object Dependencies {
     "io.circe" %% "circe-parser"
   ).map(_ % circeVersion)
 
-  val rootDependencies = Seq(
+  val rootDependencies = circe ++ Seq(
     // proactively pull in latest versions of Jackson libs, instead of relying on the versions
     // specified as transitive dependencies, due to OWASP DependencyCheck warnings for earlier versions.
     jacksonAnnotations,
@@ -121,5 +121,5 @@ object Dependencies {
     workbenchGoogleTests,
     workbenchMetrics,
     sam
-  ) ++ circe
+  )
 }
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -9,6 +9,8 @@ object Dependencies {
   val postgresDriverV = "42.2.4"
   val socketFactoryV  = "1.0.10"
   val dbcpV           = "2.5.0"
+  val sttpV           = "1.3.1"
+  val circeVersion    = "0.9.3"
 
   val workbenchUtilV    = "0.3-0e9d080"
   val workbenchModelV   = "0.11-2ce3359"
@@ -68,6 +70,14 @@ object Dependencies {
   val postgresDriver: ModuleID = "org.postgresql" % "postgresql" % postgresDriverV
   val socketFactory: ModuleID = "com.google.cloud.sql" % "postgres-socket-factory" % socketFactoryV
 
+  val sttp: ModuleID = "com.softwaremill.sttp" %% "core" % sttpV
+
+  val circe: Seq[ModuleID] = Seq(
+    "io.circe" %% "circe-core",
+    "io.circe" %% "circe-generic",
+    "io.circe" %% "circe-parser"
+  ).map(_ % circeVersion)
+
   val rootDependencies = Seq(
     // proactively pull in latest versions of Jackson libs, instead of relying on the versions
     // specified as transitive dependencies, due to OWASP DependencyCheck warnings for earlier versions.
@@ -83,6 +93,7 @@ object Dependencies {
     enumeratum,
 
     javaxServlet,
+    sttp,
 
     googleEndpointsFramework,
     googleEndpointsManagementControl,
@@ -108,5 +119,5 @@ object Dependencies {
     workbenchGoogleTests,
     workbenchMetrics,
     sam
-  )
+  ) ++ circe
 }
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -50,8 +50,9 @@ object Dependencies {
   val googleSourceRepositories: ModuleID = "com.google.apis" % "google-api-services-sourcerepo" % s"v1-rev21-$googleV" excludeAll(excludeGuavaJDK5)
 
 
-  val scalaTest: ModuleID = "org.scalatest" %% "scalatest"    % scalaTestV % "test"
-  val mockito: ModuleID =   "org.mockito"    % "mockito-core" % "2.18.3"   % "test"
+  val scalaTest: ModuleID =  "org.scalatest"   %% "scalatest"       % scalaTestV % "test"
+  val mockito: ModuleID =    "org.mockito"     % "mockito-core"     % "2.18.3"   % "test"
+  val mockServer: ModuleID = "org.mock-server" % "mockserver-netty" % "5.4.1"    % "test"
 
   // Exclude workbench-libs transitive dependencies so we can control the library versions individually.
   // workbench-google pulls in workbench-{util, model, metrics} and workbench-metrics pulls in workbench-util.
@@ -106,6 +107,7 @@ object Dependencies {
     googleSourceRepositories,
 
     scalaTest,
+    mockServer,
 
     slick,
     postgresDriver,
--- a/src/main/scala/org/broadinstitute/dsde/workbench/avram/dataaccess/HttpSamDao.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/avram/dataaccess/HttpSamDao.scala
@@ -1,0 +1,32 @@
+package org.broadinstitute.dsde.workbench.avram.dataaccess
+
+import java.util.logging.Logger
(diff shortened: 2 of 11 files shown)


project/Dependencies.scala
```diff
@@ -108,5 +121,5 @@ object Dependencies {
     workbenchGoogleTests,
     workbenchMetrics,
     sam
-  )
+  ) ++ circe
```
breilly2: I'm not sure I like breaking convention like this, but I also like how the dependencies are grouped together and the version is listed once. (That came straight from the circe docs.) Any suggestions for improvements? My only thought just now is putting `circe` first, i.e.:
```
val rootDependencies = circe ++ Seq(
  ...
)
```
Maybe there are other groupings that could be treated the same way.

src/main/scala/org/broadinstitute/dsde/workbench/avram/dataaccess/HttpSamDao.scala
```diff
@@ -0,0 +1,32 @@
...
+import io.circe.Decoder.Result
+import io.circe.{DecodingFailure, Json, ParsingFailure}
+import io.circe.generic.auto._
+import io.circe.parser._
+import org.broadinstitute.dsde.workbench.avram.util.{ErrorResponse, RestClient}
+
+class HttpSamDao(samUrl: String) extends SamDao with RestClient {
+  private val log: Logger = Logger.getLogger(getClass.getName)
+
+  override def getUserStatus(token: String): Either[ErrorResponse, SamUserInfoResponse] = {
```
breilly2, line 17: I still need to unify the error representations as suggested in #7.

ahaessly (approved):
👍

src/main/scala/org/broadinstitute/dsde/workbench/avram/dataaccess/HttpSamDao.scala
```diff
@@ -0,0 +1,32 @@
...
+    val request = buildAuthenticatedGetRequest(samUrl, "/register/user/v2/self/info", token)
+    val response: Id[Response[String]] = request.send()
+    response.body match {
+      case Left(error: String) => Left(ErrorResponse(response.code, error))
```
rtitle, line 33: I'm wondering if Avram should define its own exceptions+messages instead of bubbling up underlying error messages to the user. The Sam one in particular might be pretty verbose containing a JSON stack trace which the user doesn't need to see.

Leo does this by defining a `LeoException` trait and mapping underlying errors to a `LeoException` implementation with a user friendly status code and message. 

If you map all errors to `AvramException` it might help with monadic use of `Either`, too.

src/main/scala/org/broadinstitute/dsde/workbench/avram/util/ErrorResponse.scala
```diff
@@ -0,0 +1,36 @@
...
+  * @param message    error message
+  */
+case class ErrorResponse(statusCode: Int, message: String) {
+  private val log = Logger.getLogger(getClass.getName)
+
+  /**
+    * Create a Google Cloud Endpoints Framework exception matching the status code.
+    */
+  def exception: Exception = {
+    log.info(message)
```
rtitle: `log.error`?

src/main/scala/org/broadinstitute/dsde/workbench/avram/util/RestClient.scala
```diff
@@ -0,0 +1,27 @@
...
+
+import com.softwaremill.sttp.{HttpURLConnectionBackend, Request, Uri, sttp}
+
+/**
+  * Helpful stuff for making rest calls using sttp (https://github.com/softwaremill/sttp).
+  */
+trait RestClient {
+  private val log: Logger = Logger.getLogger(getClass.getName)
+
+  implicit val sttpBackend = HttpURLConnectionBackend() // TODO: consider refactoring to allow use of SttpBackendStub in tests
```
rtitle: Is it ok to have a shared backend for multiple request threads? In other words should this be a `def`?

rtitle (approved):

src/main/scala/org/broadinstitute/dsde/workbench/avram/dataaccess/HttpSamDao.scala
breilly2: I'll have a look at `LeoException`. Good point about immediately translating external errors into a single error catcher type potentially helping with `Either`.

src/main/scala/org/broadinstitute/dsde/workbench/avram/util/ErrorResponse.scala
breilly2: Either that or this goes away. This makes an exception that will be thrown so we may have all the logging we need already. I probably put this in so I could see the message and forgot to remove it.

I might want to change the `log.info` below to `log.warning`. If we encounter unexpected status codes, we probably want to make a decision about how to handle them.

src/main/scala/org/broadinstitute/dsde/workbench/avram/util/RestClient.scala
breilly2: Is there just one instance? My java background makes me think of a trait as a fancy interface, where nothing actually exists until an object or class instance extends the trait. Further, I can't access `RestClient.sttpBackend` directly as I would expect to be able to if it was effectively a singleton.

I'd have to run some more experiments to prove whether or not scala is retaining a single value and giving every object a reference to it, but I hope it's not doing that. I hope my mental model isn't that far off. 😬

src/main/scala/org/broadinstitute/dsde/workbench/avram/util/ErrorResponse.scala
breilly2: In fact, I'm going to declare that this is not the place to log the message. This just creates an exception object. One would expect it wouldn't be created if it isn't going to be thrown but it's also not guaranteed. I'm going to leave it to the actual error handling code to log the error.

src/main/scala/org/broadinstitute/dsde/workbench/avram/dataaccess/HttpSamDao.scala
breilly2: Looks pretty similar to the `ErrorResponse` I created. I chose that name because I wanted it to fit with the notion that it contains an HTTP status code which was pushing the outer layer a little far into the business logic for my liking.

breilly2:
I made changes to `HttpSamDao. getUserStatus`. Please re-review that. Thanks.

src/main/scala/org/broadinstitute/dsde/workbench/avram/util/RestClient.scala
rtitle: I think it's just 1 instance because there's just 1 instance of `SamDao`, which extends this trait.

I think it's probably fine, I just like to call out cases where singletons are used concurrently since I've been burned by that before.

rtitle:
Looks great! The `leftMap` approach is much better.

:shipit:

src/main/scala/org/broadinstitute/dsde/workbench/avram/dataaccess/HttpSamDao.scala
```diff
@@ -0,0 +1,46 @@
...
+
+class HttpSamDao(samUrl: String) extends SamDao with RestClient {
+  private val log: Logger = Logger.getLogger(getClass.getName)
+
+  override def getUserStatus(token: String): Either[ErrorResponse, SamUserInfoResponse] = {
+    val request = buildAuthenticatedGetRequest(samUrl, "/register/user/v2/self/info", token)
+    for {
+      response <- request.send()
+      content <- response.body                 leftMap responseToErrorResponse(response)
+      json <- parse(content)                   leftMap circeErrorToErrorResponse
```
ahaessly, line 22: minor nit - this spacing seems strange to me.

ahaessly (approved):
👍

src/main/scala/org/broadinstitute/dsde/workbench/avram/dataaccess/HttpSamDao.scala
breilly2: Understood. It's inspired by what Rob did in #12. I'm going to keep it for now but we can reevaluate.

src/main/scala/org/broadinstitute/dsde/workbench/avram/util/RestClient.scala
breilly2: Ah, you're right. I must be thinking of some other context (maybe Akka land?). I'll make sure HttpURLConnectionBackend is thread-safe.

breilly2: Update, I had looked into this before and concluded that it is thread-safe, but I still changed it to a `def` because I don't see any downside.

PR #12, opened 2018-09-27
[Experimental] FP refactor of Avram code

rtitle:
Note this branch is based off Brian's [br_sam_dao](https://github.com/DataBiosphere/avram/pull/11) branch.

This is a stab I took to implement Avram code based on a monad transformer stack. The general approach is:
1. Create a `AvramResult[A]` type which is a "stack" of the monads `Reader`, `IO`, `Either`.
2. Create utility methods to lift various combinations of these types into an `AvramResult`
3. All business logic is basically implemented as `for`-comprehensions of `AvramResult`s
4. Before the HTTP response is committed, we "unwrap" the effects to get a type `A`

More comments inline. Unit tests pass locally for me - however I haven't run this on a real environment.

I am also vaguely aware that recent developments in the Scala FP world are moving _away_ from traditional monad transformers. Examples: mtl, eff, tagless final. I'm less familiar with these libraries, but might be worth exploring as an alternative.

In all cases:


src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
```diff
@@ -21,33 +22,63 @@ class AvramRoutes {
...
   // TODO: move/merge this endpoint into a status API
   @ApiMethod(name = "dbPoolStats", httpMethod = "get", path = "dbPoolStats")
   def dbPoolStats: DbPoolStats = {
-    val dataSource = Avram.dbcpDataSource
-    val result = for {
-      totalConnections <- database.run(
-        sql"select count(*) from pg_stat_activity where pid <> pg_backend_pid() and usename = current_user".as[Int])
-    } yield DbPoolStats(dataSource.getNumActive, dataSource.getNumIdle, totalConnections.head)
-    Await.result(result, Duration.Inf)
+    val transformed = for {
```
rtitle, line 56: This illustrates the general pattern for an api method:
```
def api: Foo = {
  val transformed = for {
    a <- someAvramResult
    b <- anotherAvramResult
  } yield Foo(a, b)

  unsafeRun(transformed)
}
```

src/main/scala/org/broadinstitute/dsde/workbench/avram/dataaccess/HttpSamDao.scala
```diff
@@ -2,31 +2,25 @@ package org.broadinstitute.dsde.workbench.avram.dataaccess
...
-          case Right(json: Json) =>
-            val result: Result[SamUserInfoResponse] = json.as[SamUserInfoResponse]
-            result match {
-              case Left(error: DecodingFailure) => Left(ErrorResponse(500, error.message))
-              case Right(userInfo: SamUserInfoResponse) => Right(userInfo)
-            }
-        }
-    }
+    for {
+      response <- request.send()                        |> ioToResult
```
rtitle, line 19: Illustrates the user of utility methods to "lift" types into `AvramResult`.

Note: `|>` (or `thrush` alternatively) is just function application. E.g. `foo(3)` is equivalent to `3 |> foo`. This is just syntax, not essential for core functionality.

I thought this format was a little easier to read than `ioToResult(request.send())`. I like how the left side is cleaner and the types are on the right almost like documentation.

src/main/scala/org/broadinstitute/dsde/workbench/avram/dependencies/AvramDependencies.scala
```diff
@@ -0,0 +1,9 @@
+package org.broadinstitute.dsde.workbench.avram.dependencies
+
+import org.apache.commons.dbcp2.BasicDataSource
+import org.broadinstitute.dsde.workbench.avram.dataaccess.SamDao
+import slick.jdbc.PostgresProfile.api._
+
+case class AvramDependencies(database: Database,
```
rtitle, line 7: Container for all dependencies to "inject" into the `Reader` portion of `AvramResult`.

src/main/scala/org/broadinstitute/dsde/workbench/avram/dependencies/Global.scala
```diff
@@ -0,0 +1,14 @@
+package org.broadinstitute.dsde.workbench.avram.dependencies
+
+import org.broadinstitute.dsde.workbench.avram.config.AvramConfig
+import org.broadinstitute.dsde.workbench.avram.dataaccess.HttpSamDao
+import org.broadinstitute.dsde.workbench.avram.util.SlickDatabaseFactory
+
+object Global {
+  private val databaseFactory = new SlickDatabaseFactory(AvramConfig.dbcpDataSourceConfig)
+
+  val dependencies = AvramDependencies(
```
rtitle, line 10: Global instantiation of `AvramDependencies`

src/main/scala/org/broadinstitute/dsde/workbench/avram/util/AvramError.scala
```diff
@@ -0,0 +1,46 @@
...
+      case _ =>
+        new InternalServerErrorException(message)
+    }
+  }
+
+}
+
+object AvramError {
+  // Typeclass for converting types to AvramError
+  trait AvramErrorable[A] {
```
rtitle, line 27: This lets us implicitly convert various exception types into an `AvramError`. 

Still not certain whether this is a good idea. The alternative is to require converting errors to `AvramError` at the place of use.

Maybe there is some middle ground: we could auto-convert common circe errors but not DB errors for example.

src/main/scala/org/broadinstitute/dsde/workbench/avram/util/RestClient.scala
```diff
@@ -3,15 +3,19 @@ package org.broadinstitute.dsde.workbench.avram.util
...
 /**
   * Helpful stuff for making rest calls using sttp (https://github.com/softwaremill/sttp).
   */
 trait RestClient {
   private val log: Logger = Logger.getLogger(getClass.getName)
 
-  implicit val sttpBackend = HttpURLConnectionBackend() // TODO: consider refactoring to allow use of SttpBackendStub in tests
+  //implicit val sttpBackend = HttpURLConnectionBackend() // TODO: consider refactoring to allow use of SttpBackendStub in tests
+
+  implicit val sttpBackend = AsyncHttpClientCatsBackend[IO]()
```
rtitle, line 18: Makes sttp return an `IO[A]` which fits into our monad stack instead of an `Id[A]`.

src/main/scala/org/broadinstitute/dsde/workbench/avram/util/transformers.scala
```diff
@@ -0,0 +1,53 @@
...
+    EitherT.liftF(ReaderT(readerIO))
+  }
+
+  def futureToIO[A](future: Future[A]): IO[A] = {
+    IO.fromFuture(IO(future))
+  }
+
+  // AAAAHHHH
+  def unsafeRun[A](result: AvramResult[A], deps: AvramDependencies = Global.dependencies): A = {
+    result.value.run(deps).unsafeRunSync().fold(error => throw error.toServiceException, identity)
```
rtitle, line 50: Injects the dependency into `Reader`, runs the `IO`, and folds the `Either`

src/main/scala/org/broadinstitute/dsde/workbench/avram/util/transformers.scala
```diff
@@ -0,0 +1,53 @@
...
+  */
+object transformers {
+  // AvramResult is a stack of the effects:
+  // - Reader
+  //   - IO
+  //     - Either
+  type AvramReader[A] = ReaderT[IO, AvramDependencies, A]
+  type AvramResult[A] = EitherT[AvramReader, AvramError, A]
+
+  // Conversions to AvramResult
```
rtitle, line 25: Here are all the conversions using cats APIs on the underlying monad transformers.

src/main/scala/org/broadinstitute/dsde/workbench/avram/dataaccess/HttpSamDao.scala
breilly2: Interesting. There is definitely some appeal to how it reads. I'd love to not have to do an explicit conversion here but I guess there isn't any way around that. Since I did not know this syntax, I was going to write a bunch of functions, e.g. (types omitted because, frankly, I'd need IntelliJ to tell me what they are):
```
def sendRequest(request): = {
  response = request.send()
  ioToResult(response)
}
```
OTOH, that's a bit of extra boilerplate code. I might prefer the manual formatting chore over having to write all of those functions.

Side note related to slack discussions about auto-formatting: I wonder if scalafmt would respect the desire for the `|>`s to line up vertically.

src/main/scala/org/broadinstitute/dsde/workbench/avram/dataaccess/HttpSamDao.scala
```diff
@@ -2,31 +2,25 @@ package org.broadinstitute.dsde.workbench.avram.dataaccess
...
+      content  <- response.body.leftMap(msg =>
```
breilly2, line 20: Ahh, that's what `leftMap` is good for! Again, I was falling back on extracting functions to contain the verbosity of pattern matching (and hating the silliness of the `Right` case).

src/main/scala/org/broadinstitute/dsde/workbench/avram/dataaccess/HttpSamDao.scala
```diff
@@ -2,31 +2,25 @@ package org.broadinstitute.dsde.workbench.avram.dataaccess
...
+                    AvramError(response.code, msg))     |> eitherToResult[AvramError, String]
```
breilly2, line 21: `eitherToResult` with the explicit types is kinda wordy. Any way this could be done implicitly like you did with `AvramErrorable`? It could also be that it's useful documentation to have the types explicitly stated... I haven't decided yet.

src/main/scala/org/broadinstitute/dsde/workbench/avram/util/transformers.scala
```diff
@@ -0,0 +1,53 @@
...
+  def ioToResult[A](io: IO[A]): AvramResult[A] = {
+    EitherT.liftF(ReaderT.liftF(io))
+  }
+
+  def eitherToResult[A: AvramErrorable, B](either: Either[A, B]): AvramResult[B] = {
+    val withAvramError = either.leftMap(a => implicitly[AvramErrorable[A]].toAvramError(a))
+    EitherT.fromEither[AvramReader](withAvramError)
+  }
+
+  def withDependencies[A](reader: AvramDependencies => A): AvramResult[A] = {
```
breilly2, line 36: This makes sense. I was having a heck of a time trying to figure out how this should work.

src/main/scala/org/broadinstitute/dsde/workbench/avram/util/transformers.scala
```diff
@@ -0,0 +1,53 @@
...
+
+  def withDependenciesIO[A](readerIO: AvramDependencies => IO[A]): AvramResult[A] = {
+    EitherT.liftF(ReaderT(readerIO))
+  }
+
+  def futureToIO[A](future: Future[A]): IO[A] = {
+    IO.fromFuture(IO(future))
+  }
+
+  // AAAAHHHH
```
breilly2, line 48: 😱

src/main/scala/org/broadinstitute/dsde/workbench/avram/util/transformers.scala
breilly2: Nit: I'd put this closer to the API endpoints, maybe in a supertype, to avoid the temptation to call it from anywhere else.

breilly2:
This is a great experiment. Thank you for contributing this! I'm both impressed that you were able to put this together in an evening and discouraged that, despite much effort, I'm still very far away from being able to write this code.

Now that I see it, I'm definitely worried that deciphering compilation errors will require an understanding of the mechanics. Maybe there exists some simple wording that makes it easy for anyone to internalize the essentials but I haven't come across it yet. :( I think I'd be overly optimistic to think that I could come up with that wording. :( :(

Still, this definitely has merit and is worth considering.

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
breilly2: `unsafeRun` doesn't need any other parameters because they'd be bundled up in `transformed`, right? I'm trying to keep in mind the notion that we're first building the program to run (represented by `transformed`) and then running it.

rtitle: Correct, `transformed` is a _description_ of the program in terms of the effect stack, and `unsafeRun` actually runs it.

src/main/scala/org/broadinstitute/dsde/workbench/avram/dataaccess/HttpSamDao.scala
rtitle: It _might_ be possible to make `ioToResult` an implicit conversion so you wouldn't need to specify it at all. I'm wary however of implicit conversions because it can make the code feel too magical. Maybe I'll see if that even works..

src/main/scala/org/broadinstitute/dsde/workbench/avram/util/transformers.scala
rtitle: That's a good idea

src/main/scala/org/broadinstitute/dsde/workbench/avram/dataaccess/HttpSamDao.scala
rtitle: Yeah me neither. :) It may be possible to do it implicitly. I also kind of like the types stated.

kcibul:
Hey Brian -- that's a really important comment you made there,.  It's very important that our code is able to be understood and be maintained not only by experts, but also possible for new people to reason about and ramp up on quickly.  I'd be concerned about an approach that a PSE on the team finds complex and hard to understand.

src/main/scala/org/broadinstitute/dsde/workbench/avram/util/transformers.scala
breilly2: Long-delayed follow-up: I also instinctively put `unsafeRun` here when applying this pattern. The other place I found I want to call `unsafeRun` is from tests. Hopefully, having "unsafe" in the name is enough of a warning to be careful when calling it.

PR #13, opened 2018-10-02
GAWB-3765: Added auth endpoint

breilly2:
I'd still like to refactor the endpoint based on feedback in #7 but I have other work I need to prioritize first.

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
@@ -35,7 +35,6 @@ object PongService {
 class AvramRoutes extends BaseEndpoint {
 
   private val log = Logger.getLogger(getClass.getName)
-  private val samDao = Avram.samDao
   private val database = Avram.database
 
   @ApiMethod(name = "ping", httpMethod = "get", path = "ping")
@@ -64,12 +63,6 @@ class AvramRoutes extends BaseEndpoint {
         sql"select count(*) from pg_stat_activity where pid <> pg_backend_pid() and usename = current_user".as[Int])
     } yield DbPoolStats(dataSource.getNumActive, dataSource.getNumIdle, totalConnections.head)
     Await.result(result, Duration.Inf)
-  }
-
-  private def getToken(req: HttpServletRequest) = {
-    Option(req.getHeader("Authorization")) map {
-      _.stripPrefix("Bearer ")
-    }
   }
 
   private def fetchTimestampFromDBWithSlick(): String = {
--- a/src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
@@ -3,7 +3,6 @@ package org.broadinstitute.dsde.workbench.avram.api
 import java.util.logging.Logger
 
 import com.google.api.server.spi.config.{Api, ApiMethod}
-import com.google.api.server.spi.response.UnauthorizedException
 import javax.servlet.http.HttpServletRequest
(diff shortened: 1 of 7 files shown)


ahaessly (approved):
👍

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
```diff
@@ -46,6 +66,12 @@ class AvramRoutes {
     Await.result(result, Duration.Inf)
   }
 
+  private def getToken(req: HttpServletRequest) = {
```
breilly2: Oops... this is unused. I meant to remove this.

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
```diff
@@ -3,31 +3,51 @@ package org.broadinstitute.dsde.workbench.avram.api
...
+  }
+}
+
 @Api(name = "avram", version = "v1", scopes = Array("https://www.googleapis.com/auth/userinfo.email"))
-class AvramRoutes {
+class AvramRoutes extends BaseEndpoint {
 
   private val log = Logger.getLogger(getClass.getName)
-
+  private val samDao = Avram.samDao
```
rtitle: I don't think `samDao` is used in this class (just in the superclass)?

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramRoutes.scala
rtitle: +1 removing

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/BaseEndpoint.scala
```diff
@@ -0,0 +1,40 @@
...
+    for {
+      token <- extractBearerToken(r)
+      userInfo <- samDao.getUserStatus(token)
+    } yield userInfo
+  }
+
+  private def extractBearerToken(r: HttpServletRequest): Either[ErrorResponse, String] = {
+    val token = for {
+      value <- Option(r.getHeader("Authorization"))
+      tokenMatch <- """(?i)bearer (.*)""".r.findPrefixMatchOf(value)
```
rtitle: It may be good to define the pattern in a class-level `val` to avoid recompiling it each time:
```
val bearerPattern = """(?i)bearer (.*)""".r
```

src/test/scala/org/broadinstitute/dsde/workbench/avram/api/FakeHttpServletRequest.scala
```diff
@@ -0,0 +1,146 @@
...
+import scala.util.Try
+
+/**
+  * Stub implementations of header functions for HttpServletRequest.
+  */
+trait StubbedHeaders {
+  def headers: Map[String, List[String]]
+
+  def getHeader(name: String): String = {
+    Try(headers.filterKeys(_.equalsIgnoreCase(name))
```
rtitle: Potential alternative:
```
headers.filterKeys(_.equalsIgnoreCase(name)).values.flatten.headOption.getOrElse(null)
```

rtitle (approved):
Looks good, just minor comments

src/test/scala/org/broadinstitute/dsde/workbench/avram/api/FakeHttpServletRequest.scala
breilly2: Huh... I don't even remember writing that. I wonder if I copied it from stackoverflow or something. Anyway, IntelliJ suggested a change to yours of using `orNull` instead of `getOrElse(null)`.

PR #15, opened 2018-10-09
Add git-secrets

breilly2:
This is copied from leonardo and is the same as other projects have. (Leonardo was a good one to copy because, like avram, it does not have an rsync startup script.)

In all cases:


ahaessly (approved):
👍

ansingh7115 (approved):

PR #17, opened 2018-10-16
GAWB-3855 Changes to deploy.sh to support manually triggered indie deploys

gpolumbo-broad:
Updating deploy.sh to optionally accept the target environment as the 3rd parameter, or it will derive the target environment based on the branch name when a 3rd argument is not provided

In all cases:

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -4,20 +4,35 @@ set -x
 
 VAULT_TOKEN=$1
 GIT_BRANCH=$2
+TARGET_ENV=$3
 
-#need to get the environment from the branch name
-if [ "$GIT_BRANCH" == "develop" ]; then
-	ENVIRONMENT="dev"
-elif [ "$GIT_BRANCH" == "alpha" ]; then
-    ENVIRONMENT="alpha"
-elif [ "$GIT_BRANCH" == "staging" ]; then
-	ENVIRONMENT="staging"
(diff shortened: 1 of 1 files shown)


ansingh7115 (approved):

deploy.sh
```diff
@@ -4,21 +4,36 @@ set -x
...
+        exit 1
+    fi
+fi
+
+if [[ "$TARGET_ENV" =~ ^(dev|alpha|staging|prod)$ ]]; then
+    ENVIRONMENT=${TARGET_ENV}
 else
-	echo "Unknown Git branch $GIT_BRANCH"
-	#exit 1
+    echo "Unknown environment: $TARGET_ENV - must be one of [dev, alpha, staging, prod]"
```
ahaessly, line 30: so we also have a perf deploy?

ahaessly (commented):
👍

deploy.sh
gpolumbo-broad: @ahaessly not sure if you have a `perf` environment set up or not.  To add it, it is pretty easy, we just need to update your deploy script here, and then also update https://github.com/broadinstitute/dsp-jenkins/blob/gpolumbo-GAWB-3855/src/main/resources/IndieDeployConfig.conf to add the `perf` env as a viable target.

PR #18, opened 2018-10-22
GAWB-3753: Remove Cloud Endpoints and add Java Servlets with POST and GET Collections

ansingh7115:

In all cases:


ansingh7115:
This is a preliminary PR and NOT ready for full review yet - just want to get early feedback on something first

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramServlet.scala
```diff
@@ -0,0 +1,67 @@
...
+
+  def handleAuthenticatedRequest[T]
+  (request: HttpServletRequest, response: HttpServletResponse)
+  (f: SamUserInfoResponse => Future[Either[AvramException, T]]): T = {
+    val thing = for {
+      userInfo <- extractUserInfo(request)
+      result <- f(userInfo)
+    } yield userInfo
+    unsafeRun(Await.result(f(thing), Duration.apply(30, "second")))
+  }
```
ansingh7115, line 61: @breilly2 Changed this to take in a Future but I'm having a little trouble wrapping my head around executing the Future and the Either[Exception, T] format and how best to structure it so that it does what I want. Possibly better as a face to face conversation

project/Dependencies.scala
```diff
@@ -37,7 +37,9 @@ object Dependencies {
   val cats: ModuleID =           "org.typelevel"              %% "cats"            % "0.9.0"
   val enumeratum: ModuleID =     "com.beachape"               %% "enumeratum"      % "1.5.13"
 
-  val javaxServlet = "javax.servlet" % "servlet-api" % "2.5" % "provided"
+  //val javaxServlet = "javax.servlet" % "servlet-api" % "2.5" % "provided"
+  val javaxServlet2 = "javax.servlet" % "javax.servlet-api" % "4.0.1" % "provided"
```
breilly2: I mentioned this in person but I think we should limit ourselves to servlet 3.1. GAE says they support 2.5 and 3.1: https://cloud.google.com/appengine/docs/standard/java/runtime-java8#enhancements_and_upgrades_in_the_java_8_runtime

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/CollectionsServlet.scala
```diff
@@ -0,0 +1,72 @@
...
+import scala.concurrent.Future
+import scala.util.{Failure, Success}
+
+
+class CollectionsServlet extends HttpServlet with AvramServlet {
+
+  private val log = Logger.getLogger(getClass.getName)
+  private val collectionsService = new CollectionsService()
+
+  override def doGet(request: HttpServletRequest, response: HttpServletResponse) = {
```
breilly2, line 16: Before we get too far, I'd be interested to see if we can use a JAX-RS framework like [Jersey](https://jersey.github.io/). That should give us a nice way of specifying endpoints and extracting parameters from the path and maybe even serialization/deserialization (not sure if we'd be able to have it use circe).

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramServlet.scala
```diff
@@ -0,0 +1,67 @@
...
+  def registerOnComplete[T]
+  (future: Future[T], request: HttpServletRequest, response: HttpServletResponse)
+  (f: T => Unit): Future[T] = {
+    future onComplete {
+      case Success(result) => f(result)
+      case Failure(e: AvramException) => {
+        response.setStatus(e.status)
+        response.getWriter.write(e.message)
+      }
+      case Failure(e) => response.getWriter.write(e.getMessage)
```
breilly2: This should set the status to 500.

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramServlet.scala
```diff
@@ -0,0 +1,67 @@
...
+    future
+  }
+
+
+  def handleAuthenticatedRequest[T]
+  (request: HttpServletRequest, response: HttpServletResponse)
+  (f: SamUserInfoResponse => Future[Either[AvramException, T]]): T = {
+    val thing = for {
+      userInfo <- extractUserInfo(request)
+      result <- f(userInfo)
```
breilly2: This got a little bit jumbled so that `f()` is being invoked twice. The result of this invocation is being thrown away.

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramServlet.scala
```diff
@@ -0,0 +1,67 @@
...
+      userInfo <- samDao.getUserStatus(token)
+    } yield userInfo
+  }
+
+  private def extractBearerToken(r: HttpServletRequest): Either[AvramException, String] = {
+    val token = for {
+      value <- Option(r.getHeader("Authorization"))
+      tokenMatch <- bearerPattern.findPrefixMatchOf(value)
+    } yield tokenMatch.group(1)
+    token.toRight(AvramException(HttpServletResponse.SC_NOT_FOUND, "Auth token not found."))
```
breilly2: 🚧 This should return 401: Unauthorized.

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramServlet.scala
breilly2: `unsafeRun` should wrap anything that causes an effect (including fetching user info from Sam), ultimately leading to an `Either[AvramException, T]`. `f` is a function from `SamUserInfoResponse` to `Future[Either[AvramException, T]]`. All that's needed is to get the user info, invoke the function that returns a future, and await the result and we'll end up with an `Either[AvramException, T]`. So:
```
    unsafeRun {
      for {
        userInfo <- extractUserInfo(request)
        result <- Await.result(f(userInfo), Duration.apply(30, "second"))
      } yield result
    }
```
Though now that I look at it, `unsafeRun` isn't a good name for this function when not using something like cats `IO`. It's really just `eitherToResultOrThrow`. I should have renamed it or inlined the code.

It was easier for me to give an initial written reply but I'm happy to talk face-to-face too if that helps.

src/main/scala/org/broadinstitute/dsde/workbench/avram/dataaccess/HttpSamDao.scala
```diff
@@ -24,23 +23,23 @@ class HttpSamDao(samUrl: String) extends SamDao with RestClient {
     } yield userInfo
   }
 
-  private def responseToErrorResponse(response: Response[String]): String => ErrorResponse = error => ErrorResponse(response.code, error)
-  private def circeErrorToErrorResponse(e: io.circe.Error): ErrorResponse = ErrorResponse(500, e.getMessage)
+  private def responseToErrorResponse(response: Response[String]): String => AvramException = error => AvramException(response.code, error)
+  private def circeErrorToErrorResponse(e: io.circe.Error): AvramException = AvramException(500, e.getMessage)
```
ahaessly, line 27: i don't like losing track of the stack trace. can we also include the error itself? or at a minimum log it.

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramServlet.scala
ahaessly: i'd like to understand this more. especially if we really need the Await and the name of the method. let's talk faces

build.sbt
```diff
@@ -13,9 +13,7 @@ enablePlugins(AppenginePlugin)
 javaOptions in appengineDevServer in Compile -= (javaOptions in appengineDevServer in Compile).value.find(_.matches("-javaagent:.*"))
 javaOptions in appengineDevServer in Compile ++= Seq(
   "-Duse_jetty9_runtime=true",
-  "-D--enable_all_permissions=true" //,
-  //"-javaagent:/Users/ansingh/appengine-java-sdk-1.9.64/lib/agent/appengine-agent.jar"
-)
+  "-D--enable_all_permissions=true")
```
breilly2: What does `enable_all_permissions=true` do?

project/Dependencies.scala
```diff
@@ -37,7 +37,8 @@ object Dependencies {
   val cats: ModuleID =           "org.typelevel"              %% "cats"            % "0.9.0"
   val enumeratum: ModuleID =     "com.beachape"               %% "enumeratum"      % "1.5.13"
 
-  val javaxServlet = "javax.servlet" % "servlet-api" % "2.5" % "provided"
+  val javaxServlet = "javax.servlet" % "javax.servlet-api" % "3.1.0" % "provided"
+  val javaxWsRS = "javax.ws.rs" % "javax.ws.rs-api" % "2.0"
```
rtitle: Are we using `javax.ws.rs` (JAX-RS) currently? If not, maybe remove it from `Dependencies` until we decide to use it.

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramServlet.scala
```diff
@@ -0,0 +1,78 @@
...
+  private val samDao = Avram.samDao
+  private val bearerPattern = """(?i)bearer (.*)""".r
+
+  def registerOnComplete[T](future: Future[T], request: HttpServletRequest, response: HttpServletResponse) = {
+    future onComplete {
+      case Success(result) => {
+        response.setStatus(HttpServletResponse.SC_OK)
+      }
+      case Failure(e: AvramException) => {
+        response.setStatus(e.status)
```
rtitle, line 37: Would be good to add a `log.error` message for both `Failure` cases.

ahaessly (approved):
looks good. just one comment about exceptions

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramServlet.scala
```diff
@@ -0,0 +1,78 @@
...
+      }
+      case Failure(e) => {
+        response.setStatus(HttpServletResponse.SC_INTERNAL_SERVER_ERROR)
+        response.getWriter.write(e.getMessage)
+      }
+    }
+    future
+  }
+
+  def handleAuthenticatedRequest[T]
```
rtitle, line 50: I'm actually not certain the structure between `handleAuthenticatedRequest` and `registerOnComplete` is correct. If we fail to resolve a token with Sam, then the `unsafeRun` here will throw an exception. I don't think this exception is caught, so would result in a 500 error to the user.

I'll think about some alternatives...

src/main/scala/org/broadinstitute/dsde/workbench/avram/dataaccess/HttpSamDao.scala
rtitle: I tend to like _not_ returning stack traces in JSON responses, but logging them in the application.

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/CollectionComponent.scala
```diff
@@ -35,10 +36,15 @@ trait CollectionComponent extends AvramComponent {
...
-    def save(externalId: UUID, samResource: String, createdBy: String): DBIO[Int] = {
+    def save(externalId: UUID, samResource: SamResource, createdBy: String)(implicit executionContext: ExecutionContext): DBIO[Collection] = {
       val now = Timestamp.from(Instant.now)
       //currently the updatedBy and updatedTimestamp are the same as created at save time
-      collectionQuery += CollectionRecord(0, externalId, samResource, createdBy, now, createdBy, now)
+      val collectionRecord = CollectionRecord(0, externalId, samResource.resourceName, createdBy, now, createdBy, now)
+      for {
+        _ <- collectionQuery += collectionRecord
+      } yield {
+        unmarshalCollection(collectionRecord)
```
rtitle, line 46: I guess this would unmarshal a `CollectionRecord` where the id is 0. Is there anyway to insert to the DB and return the id that was autogenerated? I think postgres has an `INSERT ... RETURNING ...`, else we could do a insert + select.

rtitle (changes requested):
In general looks solid, though I think the interaction between `handleAuthenticatedRequest` and `registerOnComplete` needs some work

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramServlet.scala
breilly2: Yeah, that's a good observation. I _think_ this will be sufficiently different in an _IO_ world that we're better off addressing it then, but I'm not 100% sure.

src/main/scala/org/broadinstitute/dsde/workbench/avram/db/CollectionComponent.scala
breilly2: Maybe something like this: https://stackoverflow.com/a/31448129/244191

breilly2 (approved):
One change in `AvramServlet` (comment marked with 🚧) then 👍. Nice work!

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramServlet.scala
rtitle: OK, we can table this to a post-IO world.

rtitle (approved):
Approving with the caveat of doing some refactoring after Brian's PR introduces IO.

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramServlet.scala
rtitle: What I'd love to see in a post-IO world is a structure like this (psuedocode):
```
  override def doGet(request: HttpServletRequest, response: HttpServletResponse) = {
    val json = for {
      userInfo <- authWithSam(request)
      externalId <- parseExternalIdParam(request)
      dbCollection <- collectionsService.getCollection(externalId)
    } yield dbCollection.asJson.noSpaces
    
    commitResponse(json, request, response)  // the "unsafe" part
  }
```

ansingh7115: Adding a TODO comment to this PR for this future work, which we can be iterative about.

ansingh7115:
Future work from this PR: 
1. Figure out if we want to keep using bare java servlets or a different servlet tool. 
2. Decide how we want to test our servlets (dependent on number #1)

Because these are things we want to decide, I didn't add tests for our servlets yet, but they should be added when we implement 1 and 2.

PR #19, opened 2018-10-25
GAWB-3755: Add SamDao.queryAction for checking authorization

breilly2:
This PR introduces concepts from Rob's experimental FP refactor (#12) for the new code without refactoring the existing code. I did it this way for 2 reasons:

1. to avoid including changes that are unrelated to adding authorization
2. to allow a side-by-side comparison of the code

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/avram/util/AvramResult.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/avram/util/AvramResult.scala
@@ -1,0 +1,21 @@
+package org.broadinstitute.dsde.workbench.avram.util
+
+import cats.data.EitherT
+import cats.effect.IO
+import org.broadinstitute.dsde.workbench.avram.model.AvramException
+
+// Note: I'm unsatisfied with AvramResult as a name but I can't think of anything better. Suggestions welcome. -breilly
+object AvramResult {
+
+  def pure[A](a: A): AvramResult[A] = EitherT.pure(a)
+  def fromError[A](e: AvramException): AvramResult[A] = EitherT.left(IO(e))
+  def fromIO[A](io: IO[A]): AvramResult[A] = EitherT.right(io)
+  def fromEither[A](either: Either[AvramException, A]): AvramResult[A] = EitherT.fromEither[IO](either)
+
+  /**
+    * Invoke execution of computations represented by an AvramResult and produce the result value.
+    * This will throw an exception for any errors raised by the computation.
+    */
+  def unsafeRun[A](result: AvramResult[A]): A =
+    result.value.unsafeRunSync().fold(e => throw e, identity)
+}
--- a/src/main/scala/org/broadinstitute/dsde/workbench/avram/util/RestClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/avram/util/RestClient.scala
@@ -3,7 +3,9 @@ package org.broadinstitute.dsde.workbench.avram.util
 import java.net.URI
 import java.util.logging.Logger
 
-import com.softwaremill.sttp.{HttpURLConnectionBackend, Request, Uri, sttp}
+import cats.effect.IO
+import com.softwaremill.sttp.asynchttpclient.cats.AsyncHttpClientCatsBackend
+import com.softwaremill.sttp.{HttpURLConnectionBackend, Id, Request, SttpBackend, Uri, sttp}
 
 /**
   * Helpful stuff for making rest calls using sttp (https://github.com/softwaremill/sttp).
@@ -11,7 +13,9 @@ import com.softwaremill.sttp.{HttpURLConnectionBackend, Request, Uri, sttp}
 trait RestClient {
   private val log: Logger = Logger.getLogger(getClass.getName)
 
-  implicit def sttpBackend = HttpURLConnectionBackend() // TODO: consider refactoring to allow use of SttpBackendStub in tests
+  // Temporarily don't put either of these in implicit scope so that functions can choose one
+  def sttpBackend: SttpBackend[Id, Nothing] = HttpURLConnectionBackend()
+  def catsSttpBackend: SttpBackend[IO, Nothing] = AsyncHttpClientCatsBackend[IO]()
 
   def buildAuthenticatedGetRequest(url: String, path: String, token: String): Request[String, Nothing] = {
     sttp.auth.bearer(token).get(buildUri(url, path))
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -1,6 +1,8 @@
 import sbt._
 
 object Dependencies {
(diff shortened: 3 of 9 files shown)


ansingh7115:
I'm good with using AvramResult and abstracting away dealing with Eithers and IO (especially after dealing with Futures/Eithers in my own PR)

rtitle (approved):
Looks good to me

src/main/scala/org/broadinstitute/dsde/workbench/avram/util/AvramResult.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  /**
+    * Promote an IO to an AvramResult.
+    */
+  def AvramResult[A](io: IO[A]): AvramResult[A] = EitherT.right(io)
+
+  /**
+    * Promote an Either to an AvramResult.
+    */
+  def AvramResult[A](either: Either[ErrorResponse, A]): AvramResult[A] =
```
breilly2: These `AvramResult` functions allow for `IO`-like usage:
```
AvramResult { ??? }
```
which seemed convenient. However, I don't entirely like it because it's not always clear which of these functions is being called. On the flip side, it's nice to not have to worry about which is being called.

_However_, I had to stop short of adding what I originally thought would be convenient:
```
def AvramResult[A](a: A): AvramResult[A] = EitherT.pure(a)
```
The problem with this is that you might trip over it when you really mean to be calling the version that takes an `IO`. To illustrate this, add the above method and then change `HttpSamDao.queryAction` to use `sttpBackend` instead of `catsSttpBackend`. It still compiles _and_ the tests still pass! However, even though everything is still lexically "wrapped" in `AvramResult`, it's no longer safely in an `IO`... and I can't think of a test that would expose this. Almost makes me want to add this:
```
def AvramResult[A](a: A): AvramResult[A] = ??? // something here that causes compilation failure if anything references it
```
@rtitle, do you have any ideas about how to make this better?

rtitle: `AvramResult` is just a type alias for `EitherT` - why not just make callers call `pure` when passing a pure value? This may be why `EitherT` doesn't have overloaded methods -- they make you explicitly call `.fromEither`, `.pure`, `.right`, etc depending on what type you have. It makes it harder to "lift" the wrong type to an EitherT by mistake.

Also instead of `def AvramResult` you can probably do `def apply` and still get the `AvramResult { ??? }` syntax since you're already inside an `AvramResult` object.

breilly2: `apply` doesn't seem to work... I think maybe because it's an `object` rather than a `class` or `trait`? I'm not sure. I had the same initial thought but wasn't able to get any use of `apply` to compile. I think I'd have to promote `AvramResult` to a full type to get that.

Although `AvramResult` is a type alias, it doesn't seem to get the type's functions. For example, I can't just call `AvramResult.right()`. I'm sure there's a good reason why that's the case but I'm not entirely surprised either.

Provided I put the `AvramResult` type alias in a package object, I was able to add these functions to the `AvramResult` object:
* `pure` - proxies to `EitherT.pure`
* `fromEither` - proxies to `EitherT.fromEither`
* `fromIO` - proxies to `EitherT.right`
* `fromError` - uses `EitherT.left` to make a pure-like function for errors

I think I like it. Code now reads like:
```
AvramResult.pure(theAnswer)
AvramResult.fromEither(doSomethingThatMightFail())
AvramResult.right(IO(doSomethingEffecty()))
AvramResult.fromError(ErrorResponse(500, "Boom!"))
```
What do you think of that?

breilly2: I went ahead and pushed this. Let me know what you think. It's easy enough to undo.

src/main/scala/org/broadinstitute/dsde/workbench/avram/util/RestClient.scala
```diff
@@ -3,15 +3,19 @@ package org.broadinstitute.dsde.workbench.avram.util
...
 /**
   * Helpful stuff for making rest calls using sttp (https://github.com/softwaremill/sttp).
   */
 trait RestClient {
   private val log: Logger = Logger.getLogger(getClass.getName)
 
-  implicit def sttpBackend = HttpURLConnectionBackend() // TODO: consider refactoring to allow use of SttpBackendStub in tests
+  // Temporarily don't put either of these in implicit scope so that functions can choose one
+  def sttpBackend: SttpBackend[Id, Nothing] = HttpURLConnectionBackend()
+  def catsSttpBackend: SttpBackend[IO, Nothing] = AsyncHttpClientCatsBackend[IO]()
```
ansingh7115, line 18: Question: Why do we need two and why is getUserStatus using  `HttpURLConnectionBackend` while `queryAction`is using `AsyncHttpCatsBackend`?

breilly2: Having 2 is only temporary. I did this to make the change to `IO` only apply to the new code I was adding, which should help keep the PR smaller (doesn't include refactoring existing code) and keep it more compatible with other in-flight PRs. If accepted, I would follow-on quickly with a separate PR for refactoring the existing code away from `HttpURLConnectionBackend`.

src/main/scala/org/broadinstitute/dsde/workbench/avram/util/AvramResult.scala
```diff
@@ -0,0 +1,20 @@
+package org.broadinstitute.dsde.workbench.avram.util
+
+import cats.data.EitherT
+import cats.effect.IO
+
+// Note: I'm unsatisfied with AvramResult as a name but I can't think of anything better. Suggestions welcome. -breilly
+object AvramResult {
```
ahaessly, line 8: what about EntityResult(s)?

ahaessly: hmm, but I just realized it is also the return type from the queryAction call to sam. so maybe it does need to be more generic. maybe AvramDataOrError

ahaessly (approved):
one name suggestion

src/main/scala/org/broadinstitute/dsde/workbench/avram/util/AvramResult.scala
breilly2: This is more general than that. It represents a potential result... or an error. It's a thing that you have to invoke to get that result. Until you do that, no actual computation has happened. It's essentially a program represented as a data structure that's built at runtime. It's really a specialization of cats `IO`... so something like `AvramIO` maybe.

ansingh7115 (approved):
Neat - I'm enjoying the smaller, iterative PRs :)

PR #24, opened 2018-11-05
GAWB-3967: Switch to Jersey

ansingh7115:

In all cases:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,7 +45,7 @@ jobs:
           # fallback to using the latest cache if no exact match is found
           - v1-dependencies-
 
-      - run: cat /dev/null | sbt clean compile test:compile
+      - run: cat /dev/null | sbt clean compile
 
       - save_cache:
           paths:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,7 +45,7 @@ jobs:
           # fallback to using the latest cache if no exact match is found
           - v1-dependencies-
 
-      - run: cat /dev/null | sbt clean compile | sbt test:compile
+      - run: cat /dev/null | sbt clean compile test:compile
 
       - save_cache:
           paths:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,7 +45,7 @@ jobs:
           # fallback to using the latest cache if no exact match is found
           - v1-dependencies-
 
-      - run: cat /dev/null | ls ~/.ivy2/cache && rm ~/.ivy2/cache && sbt clean compile test:compile
+      - run: cat /dev/null | ls ~/.ivy2/cache && rm -rf ~/.ivy2/cache && sbt clean compile test:compile
 
       - save_cache:
           paths:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,7 +45,7 @@ jobs:
           # fallback to using the latest cache if no exact match is found
           - v1-dependencies-
 
-      - run: cat /dev/null | sbt test:compile
+      - run: cat /dev/null | sbt clean compile | sbt test:compile
 
       - save_cache:
           paths:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,7 +45,7 @@ jobs:
           # fallback to using the latest cache if no exact match is found
           - v1-dependencies-
 
-      - run: cat /dev/null | ls ~/.ivy2/cache && rm -rf ~/.ivy2/cache && sbt clean compile test:compile
+      - run: cat /dev/null | sbt clean compile test:compile
 
       - save_cache:
           paths:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,7 +45,7 @@ jobs:
           # fallback to using the latest cache if no exact match is found
           - v1-dependencies-
 
-      - run: cat /dev/null | sbt clean compile test:compile
+      - run: cat /dev/null | sbt clean compile && sbt clean compile test:compile
 
       - save_cache:
           paths:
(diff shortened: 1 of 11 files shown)


.circleci/config.yml
```diff
@@ -45,7 +45,7 @@ jobs:
           # fallback to using the latest cache if no exact match is found
           - v1-dependencies-
 
-      - run: cat /dev/null | sbt test:compile
+      - run: cat /dev/null | sbt clean compile | sbt test:compile
```
rtitle: The double pipe probably doesn't work as you expect (and may be why the circle build failed). Try making this:
```
cat /dev/null | sbt clean compile test:compile
```

project/Dependencies.scala
```diff
@@ -42,6 +41,9 @@ object Dependencies {
   val enumeratum: ModuleID =     "com.beachape"               %% "enumeratum"      % "1.5.13"
 
   val javaxServlet: ModuleID = "javax.servlet" % "javax.servlet-api" % javaxServletV % "provided"
+  val jerseyServer: ModuleID = "org.glassfish.jersey.core" % "jersey-server" % "2.26"
```
rtitle: I'd suggest putting the jersey version in a `val` at the top.

Also I'm curious about `glassFishInject` because I don't see it used. Was it copy-paste or do we actually use it somehow?

project/PackagingTypePlugin.scala
```diff
@@ -0,0 +1,8 @@
+import sbt._
+
+object PackagingTypePlugin extends AutoPlugin {
```
rtitle, line 8: Taken from here I see :)

https://github.com/sbt/sbt/issues/3618#issuecomment-424924293

Might be nice to include the above github link as a comment to help readers of the code.

This workaround worked for me but I had to `sbt clean` first -- probably why the circle build failed, since it isn't cleaning right now as written.

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramServlet.scala
```diff
@@ -14,60 +16,58 @@ import org.broadinstitute.dsde.workbench.avram.util.AvramResult.unsafeRun
...
-    def writeBody(body: T): Unit = {
-      response.setStatus(HttpServletResponse.SC_OK)
-      response.setContentType("application/json")
-      // CORS?
-      response.getWriter.write(body.asJson.noSpaces)
+                                   (implicit encoder: Encoder[T]): Response = {
+    def writeBody(body: T): Response = {
+      Response
+        .status(Status.OK)
+        .entity(body.asJson.noSpaces)
```
rtitle, line 33: Did you lose the `application/json` content type header? Or does jersey magically add it somehow?

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramServlet.scala
```diff
@@ -14,60 +16,58 @@ import org.broadinstitute.dsde.workbench.avram.util.AvramResult.unsafeRun
...
       // Log cause. TODO: look into a logging framework on top of java.util.logging
       e.cause.foreach(t => log.log(Level.SEVERE, s"Caused by:", t))
-
-      response.setStatus(e.status)
-      response.setContentType("text/plain")
-      // CORS?
-      response.getWriter.write(e.regrets)
+      Response
+        .status(e.status)
+        .entity(e.regrets)
```
rtitle, line 44: Same as above about content-type.

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramServlet.scala
```diff
@@ -14,60 +16,58 @@ import org.broadinstitute.dsde.workbench.avram.util.AvramResult.unsafeRun
...
 
   implicit val executionContext: ExecutionContext = scala.concurrent.ExecutionContext.Implicits.global
 
   val log: Logger = Logger.getLogger(getClass.getName)
-  private val samDao = Avram.samDao
+  private val samDao = avram.samDao
   private val bearerPattern = """(?i)bearer (.*)""".r
 
-  def handleAuthenticatedRequest[T](request: HttpServletRequest, response: HttpServletResponse)
+  def handleAuthenticatedRequest[T](bearerToken: String)
```
rtitle: Maybe name this `authorizationHeader` -- we need to call `extractBearerToken` to actually get the bearer token.

rtitle:
I did a pass -- just some minor comments above, looks good in general.

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/CollectionsServlet.scala
```diff
@@ -1,54 +1,44 @@
...
 import org.broadinstitute.dsde.workbench.avram.service.CollectionsService
 import org.broadinstitute.dsde.workbench.avram.util.AvramResult
+import org.broadinstitute.dsde.workbench.avram.Avram
 
 import scala.util.Try
 
-
-class CollectionsServlet extends HttpServlet with AvramServlet {
+@Path("/api/collections/v1")
+class CollectionsServlet(avram: Avram) extends AvramServlet(avram) {
```
breilly2: This isn't technically a servlet anymore. Maybe rename these back to `*Endpoint`. Jersey/JAX-RS is a framework similar to Cloud Endpoints Framework.

breilly2 (approved):
This looks really good! I 2nd much of what Rob said but I don't have much of anything to add. I was worried that returning `Response` from the endpoint functions would make testing difficult but it looks like its easy enough to verify the returned entity and status code so 👍.

src/main/scala/org/broadinstitute/dsde/workbench/avram/api/AvramServlet.scala
ansingh7115: Added @produces annotation

PR #25, opened 2018-11-06
GAWB-3755: Added authorization checks for create/get operations

breilly2:
I included some experimenting with different styles of making the checks. Will need to choose one and clean up the other before merging this PR.

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/avram/UserInfo.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/avram/UserInfo.scala
@@ -1,0 +1,14 @@
+package org.broadinstitute.dsde.workbench.avram
+
+import org.broadinstitute.dsde.workbench.avram.dataaccess.SamUserInfoResponse
+
+case class UserInfo(subjectId: String, email: String, enabled: Boolean, token: String)
+
+object UserInfo {
+  def apply(samUserInfo: SamUserInfoResponse, token: String): UserInfo =
+    UserInfo(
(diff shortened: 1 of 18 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/avram/UserInfo.scala
```diff
@@ -0,0 +1,14 @@
...
+case class UserInfo(subjectId: String, email: String, enabled: Boolean, token: String)
+
+object UserInfo {
+  def apply(samUserInfo: SamUserInfoResponse, token: String): UserInfo =
+    UserInfo(
+      samUserInfo.userSubjectId,
+      samUserInfo.userEmail,
+      samUserInfo.enabled,
+      token)
+}
```
ansingh7115, line 14: Should this be in AvramModel.scala?

breilly2: It could, though I think of a system's model as the domain it controls... collections and entities in this case. `UserInfo` is just a convenience object to bundle the info from Sam with the token (for later queries to Sam). That said, I didn't mean to leave it in `org.broadinstitute.dsde.workbench.avram` but I also wasn't sure about putting it in `util` or making a new package. I'm open to suggestions.

ansingh7115 (approved):
Looks great - thanks for finishing out the last piece of sam auth! :)

ahaessly (approved):
👍

